PR #2, opened 2018-01-18
304-login-user-is-unable-to-login-to-csl-platform

superafroman:


superafroman:
Rebased to pull in prettier config and applied to changes.

PR #4, opened 2018-01-18
330-show-user-profile-after-login

jenbutongit:
when/if logged in, /login redirects to /profile - which shows { email address, department, profession, grade }


jenbutongit:
[Show user profile after login](https://trello.com/c/AFffDHst/330-show-user-profile-after-login)

PR #5, opened 2018-01-19
327-setup-ui-framework

tav:
This PR adds the following functionality:

* We no longer need to keep adding `<script>` blocks like the following in each of our component/page files to list all the components that are used:

```html
<script>
    import Page from '../component/Page.html';
    import Layout from '../component/Layout.html';
    import Searchbar from '../component/Searchbar.html';

    export default {
        components: { Layout, Page, Searchbar }
    };
</script>
```

* The `<script>` block is automatically generated instead.

* It expects all reusable components to be defined directly in the top-level `component` directory, and page components, which can be nested within subdirectories, to be defined within the top-level `page` directory.

* The pages can then be rendered with our `svelte.render` function defined within `ui/svelte`, e.g.

```ts
import * as svelte from 'ui/svelte'

svelte.render('sign-in', {sessionDataKey: 'abcd'})
```

*  ^ This will look for either `page/sign-in.html` or `page/sign-in/index.html` component files to exist.

* A `ui/render` module has been added where we can provide wrapper functions for rendering these pages with type annotations, e.g.

```ts
export interface Profile {
	department: string
	emailAddress: string
}

export function profile(props: Profile) {
	return svelte.render('profile', props)
}
```

* We can then just import the `render` module and call the appropriate function and benefit from type safety/autocompletion/etc., e.g.

```ts
import * as render from 'ui/render'

render.profile({department: 'Home Office', email: 'alice@example.com'})

// uh oh! that should have been emailAddress and not email!
```

* As a bonus, our tests can use these functions too, and see if they generate the expected HTML when given different input parameters.

* And, finally, in dev mode, it resets the cache of compiled template modules whenever it detects any files to have changed within the component/page directories, so you no longer need to restart the dev server when trying out various templating changes.

The bulk of the functionality is introduced by our new `ui/svelte` module in `src/svelte.ts`. I also updated the existing templates to remove their `<script>` blocks and updated the main `server.ts` file to use the `render` module.

And, oh, one more thing: various files seem to have the executable bit set on them and I had to go in and `chmod -x` them. Not sure if it's someone's IDE which is doing this, but perhaps they could look into fixing that if possible.

Cheers! 👍


tav:
[Setup UI framework](https://trello.com/c/rziiHtw4/327-setup-ui-framework)

PR #6, opened 2018-01-19
#303 user can login to csl platform

superafroman:
A few tweaks to fix login after moving to lpg.dev.cshr.digital. Not perfect, but working.


tav:
[Login: User can login to CSL platform](https://trello.com/c/Ndlgpe8k/303-login-user-can-login-to-csl-platform)

PR #14, opened 2018-01-23
feature/296-user-profile-integrate-csl-profile-ui-with-api

jenbutongit:
this PR includes
* `POST` method for `/profile` route. 
When the user submits the profile form, it will `POST` to `/profile`, which will `PUT` to WSO2.
* Successful `PUT` will firstly update req.user, and then redirect to `/profile` and render the page with the updated req.user

### DIFF
--- a/page/profile/index.html
+++ b/page/profile/index.html
@@ -1,0 +1,19 @@
+<Page title="Profile">
+    <Layout page='profile'>
+        <div class="profile-page">
+            <div class="container page">
+                <div class="row">
+                    <div class="col-md-6 offset-md-3 col-xs-12">
+
+                        <div class="column-two-thirds">
+
+                            <h1 class="heading-large">Profile</h1>
+                            <User :sessionDataKey :id :emailAddress :department :profession :grade/>
+
+                        </div>
+                    </div>
+                </div>
+            </div>
+        </div>
+    </Layout>
+</Page>
--- a/service/ui/src/controllers/user.ts
+++ b/service/ui/src/controllers/user.ts
@@ -82,13 +82,8 @@ export let updateUser = (req: Request, res: Response) => {
 		},
 	}
 
-	console.log('options: ' + JSON.stringify(options))
-
-	// res.send()
-
 	request(options, function(error: Error, response: Response, body: Body) {
 		if (!error && response.statusCode == 200) {
-			console.log(req.user)
 			let updatedUser = JSON.parse(body).CshrUser
 			updateUserObject(req, updatedUser)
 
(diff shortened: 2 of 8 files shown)


page/profile/index.html
```diff
@@ -0,0 +1,19 @@
+<Page title="Profile">
+    <Layout page='profile'>
+        <div class="profile-page">
+            <div class="container page">
+                <div class="row">
+                    <div class="col-md-6 offset-md-3 col-xs-12">
```
superafroman: Can you update this to same structure as in sign-up - i.e. remote 3 of the divs and change the remaining one ro class="grid-row"

service/ui/src/controllers/user.ts
```diff
@@ -1,6 +1,8 @@
 import * as config from 'config'
 import {Request, Response} from 'express'
 import * as template from 'ui/template'
+import * as request from 'request'
+import {updateUser} from '../../dist/controllers/user'
```
tav: This should be `ui/controllers/user`?

service/ui/src/controllers/user.ts
```diff
@@ -53,5 +55,47 @@ function renderSignIn(props: SignIn) {
 }
 
 function renderProfile(props: User) {
+	console.log(props)
```
tav: Remove the debug line?

service/ui/src/controllers/user.ts
superafroman: Should be `ui/controllers/user`

service/ui/src/controllers/user.ts
superafroman: Use log4js if you want logging in, otherwise remove.

service/ui/src/server.ts
```diff
@@ -74,6 +82,7 @@ app.all(
 		failureFlash: true,
 	}),
 	(req, res) => {
+		console.log(res.user)
```
superafroman: Remove or use log4js

service/ui/src/controllers/user.ts
```diff
@@ -53,5 +55,47 @@ function renderSignIn(props: SignIn) {
...
+	}
+
+	request(options, function(error: Error, response: Response, body: Body) {
+		if (!error && response.statusCode == 200) {
+			let updatedUser = JSON.parse(body).CshrUser
+			updateUserObject(req, updatedUser)
+
+			res.send(renderProfile(req.user))
+		} else {
+			res.send(console.log(error))
```
tav: Break this apart into two lines, and set a proper response status code?

service/ui/src/server.ts
tav: Remove this debug line too?

PR #15, opened 2018-01-23
test/278-a-learner-can-give-feedback-on-the-lpg

williampri:


williampri:
[A learner can give feedback on the LPG (1)](https://trello.com/c/l5fhFF5O/278-a-learner-can-give-feedback-on-the-lpg-1)

test/headless/src/page/profile.ts
```diff
@@ -1,4 +1,5 @@
 //profile page
-// import * as puppeteer from 'puppeteer'
 
-// export const selectors: {[key: string]: string} = {}
+export const selectors: {[key: string]: string} = {
```
tav: Instead of `{[key: string]: string}`, you may want to use the builtin `Record` type in the future, e.g.

```ts
Record<string, string>
```

It's basically a shortcut for the other.

williampri: Thanks, I'll add going forward

PR #20, opened 2018-01-24
feature/293-user-profile-update-user-profile-data-ui

superafroman:
No drop downs for profile details yet, but user journey implemented. Also includes a quick fix for logout (needs the latest wso2)


service/ui/page/profile/edit-success.html
```diff
@@ -0,0 +1,17 @@
+<Page title="Edit Profile Complete">
```
jenbutongit: add :user here, so the sign out link displays properly.

superafroman: Updated to make current user available globally so we don't have to pass around.

service/ui/component/Nav.html
```diff
@@ -5,6 +5,7 @@
 		</a>
 		<nav id="proposition-menu">
 			<ul id="proposition-links">
+				{{#if signedInUser}}
```
jenbutongit: Now only the sign in link works, is signedInUser getting passed into this component for it to use?

jenbutongit: Now only the sign in link works/ is visible, is signedInUser getting passed into this component for it to use?

PR #29, opened 2018-01-26
feature/348-implement-node-catalog-service

tav:
This PR implements the `catalog` service as a library within the `ui` service that mimics the API that will be used down the line, e.g.

```js
const resp = await catalog.search(elko.context(), {
  first: 10,
  tags: ['alice', 'david', 'tav'],
})
```

A few things to note:

* The import path for the catalog service is currently `ui/service/catalog`. Once finished, it will probably be just `service/catalog`.

* Similarly, the import path for `elko` is currently `ui/service/elko`. Once finished, it will be just `elko`.

* I decided to throw Errors instead of returning a `(response, error)` tuple as it's easier to handle for the common case.

There is a fully sequenced example added in `server.ts` as `catalogDemo`. You need `dgraph` running in order to be able to run this demo. The easiest way to do that is by running docker compose within the repo.

If, like me, you prefer docker compose to use fresh containers every time, then run:

    docker-compose up --force-recreate

### DIFF
--- a/bin/gen-service-proto
+++ b/bin/gen-service-proto
@@ -1,0 +1,49 @@
+#! /usr/bin/env python
+
+import sys
+
+def error(msg):
+    print '!! ERROR: %s' % msg
+    sys.exit(1)
+
+args = sys.argv[1:]
+if not args:
+    print 'Usage: gen-service-proto [METHODS ...]'
+    sys.exit()
+
+messages = []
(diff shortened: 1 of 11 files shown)


tav:
[Implement Node Catalog Service](https://trello.com/c/WViWAjIo/348-implement-node-catalog-service)

PR #34, opened 2018-01-31
feature/ci-build-all-services

superafroman:


superafroman:
Again, PaaS should affect this. This is also inline with what we discussed for deployment - deployment will be triggered using one commit hash for all lpg-service services. (?)

PR #37, opened 2018-01-31
feature/378-figure-out-restructure-of-repo-so-code-can-be-shared-easily-between-services

jenbutongit:
PR includes:

* lib/, lib is where we should keep common code for the services. The services can depend on `lib` relatively, and use the modules that are in there
* readme :)


superafroman:
Can't seem to reply to your comment @tav - we discussed this yesterday and agreed the repo would stay mono to aid dev. How does PaaS get in the way of this?

PR #45, opened 2018-02-05
Feature/i18n

jenbutongit:


lib/config/passport.ts
```diff
@@ -2,6 +2,7 @@ import {Request, Response, NextFunction} from 'express'
 import * as passport from 'passport'
 import {Strategy} from 'passport-saml'
 
+
```
tav: In future, could we avoid no-op commits like this?

lib/package.json
```diff
@@ -8,13 +8,15 @@
 	"devDependencies": {
 		"@types/config": "0.0.33",
 		"@types/passport": "0.4.2",
-		"@types/passport-saml": "0.15.5"
+		"@types/passport-saml": "0.15.5",
+		"@types/i18n": "^0.8.3"
```
tav: In future, could we use explicit version numbers instead of version ranges?

lib/service/translation/locales/de.json
```diff
@@ -0,0 +1,4 @@
+{
```
tav: Do we have need of a German locale or is this just for testing?

PR #52, opened 2018-02-07
feature/386-tech-for-the-formal-learning-opportunity-overview-page

superafroman:
Updated overview page so generated from course data.

Building key information from tags and new duration property on course.


service/ui/src/controllers/learning-record/index.ts
```diff
@@ -110,13 +112,29 @@ async function getCourseRecord(user: any, course: any) {
...
+function getCompletionDate(statements: [any]) {
+	if (statements.length === 0) {
+		return null
+	}
+	const completed = statements.find(
+		statement =>
+			statement.verb.id === 'http://adlnet.gov/expapi/verbs/completed'
+	)
+	if (completed) {
+		return moment(completed.timestamp).format('DD/MM/YYYY')
```
tav: For future, could we use the builtin datetime formatting instead of `moment`? e.g.

```js
new Intl.DateTimeFormat('en-GB').format(date)
```

PR #60, opened 2018-02-08
feature/380-learners-can-see-an-overview-of-a-learning-opprtunity-informal

superafroman:
Moving video work into overview page.


src/service/ui/assets/js/video.js
```diff
@@ -1,30 +1,30 @@
...
+	videoId = document.getElementById('video-id').value
+
 	window.onYouTubeIframeAPIReady = function() {
 		var player = new YT.Player('video-player', {
-			height: VIDEO_HEIGHT,
-			videoId: VIDEO_ID,
-			width: VIDEO_WIDTH,
+			height: 390,
+			videoId: videoId,
+			width: 640,
```
superafroman: Note these are the defaults, but left here to indicate how we can adjust size to match any UI designs. Don't want to adjust video based on the content as we need it to fit our screens.

PR #61, opened 2018-02-08
feature/learning-results

superafroman:


src/service/ui/page/learning-record/index.html
```diff
@@ -11,7 +11,7 @@ <h1 class="heading-large">Your learning record</h1>
 					<li>
 						{{ course.title }}
 						<span>
-							{{course.result}}
+							{{course.result.result}}
```
tav: Avoid stuttering in future if possible...

PR #63, opened 2018-02-09
feature/learning results

superafroman:
Added score and pass / fail to learning results page.

Updated basket page so required content dynamically driven from catalogue.


src/lib/ui/template.ts
```diff
@@ -66,7 +67,11 @@ function compile(
 export default {
 components: {${componentNames}},
 data() {
-    return {html: htmlModule, signedInUser: getCurrentRequest().user}
+    return {
+        html: htmlModule, 
+        signedInUser: getCurrentRequest().user,
+        i18n: i18nModule
```
tav: I think we want to pass in the `res.__.bind(res)` for `const res = getCurrentRequest()` here, and not `i18n` as the global `this.locale` won't match with the in-flight request? Just helped @jenbutongit set this up, so can replace this with that.

PR #65, opened 2018-02-12
chore/overview-cleanup

superafroman:
A little more than just the course overview...

 * Added caching to SCORM content
 * Override new tincan api wrapper to remove Authorisation header when no user/pass specified. 
 * Clean up of data in learning record page
 * i18n lookup of key attributes for course overview page.


src/service/ui/page/learning-record/index.html
```diff
@@ -6,16 +6,38 @@
...
-						</span>
-					</li>
+					<tr>
+						<td>{{ course.completionDate }}</td>
+						<td>{{ course.title }}</td>
+						<td>
+							{{#if course.result.score }}
+								{{ result.score.scaled * 100 }}%
+							{{else}}
+								{{ i18n.__(course.result.result) }}
```
tav: Does this actually work?

src/service/ui/src/server.ts
```diff
@@ -2,6 +2,7 @@ import * as bodyParser from 'body-parser'
 import * as compression from 'compression'
 import * as express from 'express'
 import * as session from 'express-session'
+import * as cache from 'express-cache-response'
```
tav: For future, imported variables should not be callable, as that's not forward-compatible with the ES Modules spec. We've made a couple of exceptions so far, e.g. `express`, should also note this exception and try and avoid them in future.

PR #67, opened 2018-02-12
bug/394-bug-csl-navigation-menu-on-mobile

jenbutongit:
structure/style is more like govuk-templates now. Menu is already expanded with js disabled

### DIFF
--- a/src/service/ui/component/Header.html
+++ b/src/service/ui/component/Header.html
@@ -2,7 +2,7 @@
 	<div class="header-wrapper">
 		<div class="header-global">
             <div class="header-logo">
-                <a href="/" title="Go to the GOV.UK homepage" id="logo" class="content">
+                <a href="/" title="Go to the Learning Platform homepage" id="logo" class="content">
                     <img src="img/gov.uk_logotype_crown_invert_trans.png" width="36" height="32" alt=""> Civil Service
                 </a>
             </div>
--- a/src/service/ui/component/Header.html
+++ b/src/service/ui/component/Header.html
@@ -1,13 +1,19 @@
 <header role="banner" id="global-header" class=" with-proposition">
 	<div class="header-wrapper">
 		<div class="header-global">
-			<div class="header-logo">
-				<a href="/" title="Go to the GOV.UK homepage" id="logo" class="content">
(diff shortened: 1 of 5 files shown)


src/service/ui/component/Header.html
```diff
@@ -1,13 +1,19 @@
...
 		<div class="header-global">
-			<div class="header-logo">
-				<a href="/" title="Go to the GOV.UK homepage" id="logo" class="content">
-					<img src="img/gov.uk_logotype_crown_invert_trans.png" width="36" height="32" alt=""> Learning Platform
-				</a>
-			</div>
-		</div>
-		<Nav />
+            <div class="header-logo">
+                <a href="/" title="Go to the GOV.UK homepage" id="logo" class="content">
```
tav: Perhaps the title should be different?

src/service/ui/component/Page.html
```diff
@@ -20,9 +20,11 @@
     <Header />
 
     <slot/>
+    <script src="./js/govuk_frontend_toolkit/govuk/show-hide-content.js"></script>
+    <script src="./js/govuk_frontend_toolkit/govuk/modules.js"></script>
     <script src="./js/main.js"></script>
-        <script src="./js/govuk_frontend_toolkit/govuk/show-hide-content.js"></script>
-        <script src="./js/govuk_frontend_toolkit/govuk/modules.js"></script>
+    <script>if (typeof window.GOVUK === 'undefined') document.body.className = document.body.className.replace('js-enabled', '');</script>
```
tav: Does this work? Doesn't svelte complain about this?

src/service/ui/component/Page.html
```diff
@@ -20,9 +20,11 @@
     <Header />
 
     <slot/>
+    <script src="./js/govuk_frontend_toolkit/govuk/show-hide-content.js"></script>
```
tav: Also, make these links absolute from the root `/` instead of relative?

src/service/ui/component/Page.html
jenbutongit: Yes it works, and no svelte doesn't complain, maybe because it's the parent? unsure 😅

tav: Yay!

PR #78, opened 2018-02-15
bug/fix-travis-sass

tav:


.travis.yml
```diff
@@ -20,6 +20,8 @@ jobs:
       - npm install
     - script:
       - npm run build
+      - npm run sass:management
+      - npm run sass:ui
```
superafroman: Why does build not include sass compilation? If we want to separate typescript compilation we should have a 'compile' step as 'build' suggests everything is done.

PR #82, opened 2018-02-15
chore/switch-to-local-host

tav:


src/lib/config/index.ts
```diff
@@ -44,11 +44,22 @@ export const LOGGING = set(
...
+export const POSTGRES = set(
+	process.env.POSTGRES || 'postgres://user:password@host:5433/lpg'
+)
+
 export const SESSION_SECRET = set(
 	process.env.SESSION_SECRET ||
 		'dcOVe-ZW3ul77l23GiQSNbTJtMRio87G2yUOUAk_otcbL3uywfyLMZ9NBmDMuuOt'
 )
 
+export const UI_SERVER = set('lpg.local.cshr.digital:3001', {
```
superafroman: These should be env variables configured on the host - we shouldn't need to change code to deploy to a new environment.

src/lib/config/index.ts
```diff
@@ -44,11 +44,22 @@ export const LOGGING = set(
 	}
 )
 
+export const POSTGRES = set(
+	process.env.POSTGRES || 'postgres://user:password@host:5433/lpg'
```
superafroman: Guessing this isn't really meant to be part of this change, but obviously `@host` won't do much.

src/management-ui/page/courses/list.html
```diff
@@ -6,22 +6,21 @@ <h1 class="heading-large">Courses</h1>
...
-                                    <!-- FIXME: not portable -->
-                                    <a href="http://lpg.dev.cshr.digital:3001/courses/{{course.uid}}/">Review</a>
-                                </td>
-                            </tr>
+                        <tr>
+                            <th scope="row">
+                                <a href="/courses/{{course.uid}}/edit">{{course.title}}</a>
+                            </th>
+                            <td>
+                                <a href="//{{config.UI_SERVER}}/courses/{{course.uid}}/">Review</a>
```
superafroman: This was a hack to let me easily check things. It won't make it into private beta and it won't work once we add auth to the management ui. If we actually want to allow preview we'd need to preview it through the management ui and block xapi events etc.. tldr: I don't think there's an actual need for configuring the UI host or necessarily exposing config to templates.

tav:
[Add support for the new .local host and the new environments](https://trello.com/c/udn1FIQF/439-add-support-for-the-new-local-host-and-the-new-environments)

PR #83, opened 2018-02-16
feature/440-setup-shared-creds-and-load-dev-environment-values-from-keybase

tav:


tav:
[Setup shared creds and load dev environment values from Keybase](https://trello.com/c/lNEyxbz4/440-setup-shared-creds-and-load-dev-environment-values-from-keybase)

PR #94, opened 2018-02-21
feature/447-get-ci-to-actually-work

tav:


environ/ci
```diff
@@ -0,0 +1,59 @@
...
+npm install
+
+stage "Building the main src"
+npm run build
+
+stage "Building the new lpg-services docker image"
+docker build --cache-from cshr/lpg-services:latest -t cshr/lpg-services:${TRAVIS_COMMIT} .
+
+stage "Pushing the lpg-services:${TRAVIS_COMMIT} docker image"
+docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD"
```
superafroman: Now it's all in one script a single login would work fine.

environ/ci
```diff
@@ -0,0 +1,59 @@
...
+docker push cshr/lpg-services:${TRAVIS_COMMIT}
```
superafroman: We should tag and push `latest` as well

PR #96, opened 2018-02-21
Feature/433 402 a learner can book onto a training course m

jenbutongit:
PR includes changes for:
[**402 A learner can see bookable courses**](https://trello.com/c/tAH8M1Bw/402-a-learner-can-see-bookable-courses-m)
* First a bookable course must be created via management ui. Tag with `area-of-work:commercial` then it will show up in the 'suggested for you' if the profile is set to the commercial profession. 

[**433 A learner can book onto a training course**](https://trello.com/c/JygbmWEH/433-a-learner-can-book-onto-a-training-course-m)
* This flow matches the current csl booking pages
* currently, these bookings are not recorded anywhere 
* on the "confirm booking page" the accessibility details change link goes to nowhere because I've not actually been given this page in the flow/screenshots I've been given ¯\_(ツ)_/¯. I can remove this if required

[**413 Purchase order po or financial approver fap validation**](413-purchase-order-po-or-financial-approver-fap-validation-m)
* this is just a conditional looking for hmrc, co or dh departments and showing the relevant input


src/ui/page/suggested.html
```diff
@@ -9,7 +9,11 @@ <h1 class="heading-large">Suggested for you</h1>
                     {{#each courses as course}}
                     <tr>
                         <td>
-                            <a href="/courses/{{ course.uid }}">{{ course.title }}</a>
+                            {{#if course.type == 'bookable'}}
```
superafroman: As discussed good to have this in the courseController like video/elearning/..

src/ui/page/booking/confirmed.html
```diff
@@ -0,0 +1,28 @@
+<Page title="Boooking confirmed">
```
superafroman: ooo 😯

src/ui/controllers/course/index.ts
```diff
@@ -55,14 +53,10 @@ function getCourseDetails(
...
-function getTagValues(
-	req: extended.CourseRequest,
-	tagName: string,
-	tags: string[]
-) {
+function getTagValues(tagName: string, tags: string[]) {
 	return tags
 		.filter(tag => tag.startsWith(tagName))
-		.map(tag => req.__(tag.replace(`${tagName}:`, '')))
+		.map(tag => i18n.__(tag.replace(`${tagName}:`, '')))
```
superafroman: Probably a rebase issue - `req.__` is the right way to do this (uses the users locale)

src/ui/controllers/booking.ts
```diff
@@ -0,0 +1,168 @@
+import * as express from 'express'
+import * as template from 'lib/ui/template'
+import * as courseController from './course/index'
+import * as catalog from 'lib/service/catalog'
+import * as model from 'lib/model'
+import * as dateTime from 'lib/datetime'
+
+export async function renderBookableCourseInformation(
```
superafroman: As discussed this will/should happen in courseController, keep this controller for the actual booking process. I'd drop the breadcrumbs from the overview page for now too as they're not on any other overview page.

src/ui/page/booking/confirmed.html
jenbutongit: ooops

src/ui/controllers/course/index.ts
jenbutongit: using current master's version of this func now

PR #98, opened 2018-02-22
bug/442-bug-the-columns-on-pages-dont-always-align-with-the-outer-layer

jenbutongit:
also moved feedback into its own component

### DIFF
--- a/src/ui/component/Layout.html
+++ b/src/ui/component/Layout.html
@@ -21,23 +21,7 @@
 	<div class="main-content">
 		<slot></slot>
 	</div>
-	<div class="column-full">
-		<div class="feedback">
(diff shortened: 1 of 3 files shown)


src/ui/component/Layout.html
```diff
@@ -53,7 +37,7 @@ <h2>Help us improve Civil Service Learning</h2>
 				</div>
 			</div>
 			<div class="copyright">
-				<a href="https://www.nationalarchives.gov.uk/information-management/re-using-public-sector-information/copyright-and-re-use/crown-copyright/">© Crown copyright</a>
```
tav: Do we have to support browsers that don't support `utf-8`? 😂

jenbutongit: #accessible :)

PR #103, opened 2018-02-22
Searching for learning

smoga-git:
Enjoy


src/bin/dg2el.ts
```diff
@@ -0,0 +1,65 @@
+import * as es from 'elasticsearch'
```
smoga-git: This is superfluous but an example of a quick and dirty way of gettting dgraph -> elasticsearch

superafroman:
@smoga-git rather than a merge can you rebase onto master?

smoga-git:
scrub this branch

PR #105, opened 2018-02-23
Feature/lpg team alerted when a booking is made or cancelled

smoga-git:
Okay this has working messaging but it requires a correct inbound webhook URI. I created a demo slack  account to test functionality


superafroman:
@tav I've rebased this now so ready for review.

src/lib/service/messaging/index.ts
```diff
@@ -0,0 +1,23 @@
...
+	return {
+		api: (message: string) => {
+			axios.post(uri, {
+				text: message,
+			})
+		},
+	}
+}
+
+export function send(message: string, dispatcher: MessageDispatcher) {
```
tav: This is somewhat confusing. Could we at some point in the future, simplify to just having a `send(msg)` function which just uses the `config.BOOKING_ALERT_WEBHOOK` url directly? Thanks!

superafroman: Yeah, it's ended up a little more complicated than it needs to be for now as I mentioned to Alan wanting to be able to support multiple endpoints - e.g. have the slack message and the rpi pinged to flash some lights 🕺 🎉

PR #109, opened 2018-02-26
feature/433 a learner can book onto a training course m

superafroman:
All of the good bookings stuff.


src/ui/controllers/booking.ts
```diff
@@ -181,19 +195,25 @@ export async function tryCancelBooking(
 	res: express.Response
 ) {
 	const course = req.course
+	const record = await learnerRecord.getCourseRecord(req.user, course)
+
+	if (!record.selectedDate) {
+		res.redirect('/')
```
tav: Is this where we want to redirect them?

src/lib/youtube.ts
```diff
@@ -11,32 +11,8 @@ export interface BasicInfo {
...
-	duration += (parseInt(parts[7], 10) || 0) * 60
-	duration += (parseInt(parts[6], 10) || 0) * 3600
-	duration += (parseInt(parts[4], 10) || 0) * 86400
-	// Accept the leading minus sign for now, but might want to abort in future.
-	if (parts[1]) {
-		return -duration
-	}
-	return duration
+function convertDuration(isoDuration: string): string {
+	return isoDuration.substr(2).toLowerCase()
```
tav: Is this format good enough?

src/ui/component/Learning.html
```diff
@@ -11,13 +11,17 @@
                 </a>
             {{/if}}
             {{#if course.state}}
-                {{#if course.type === 'elearning'}}
-                <a href="/courses/{{course.uid}}/do">Resume</a>
+                {{#if course.type === 'face-to-face'}}
```
tav: Are we not calling this `event` like yuksel proposed?

superafroman: The course is a face to face course which has events - had a follow on discussion around this with the business last week.

src/lib/youtube.ts
superafroman: Yeah, it's consistent with how they label the other course durations.

src/ui/controllers/booking.ts
superafroman: Ideally we'd send them some malware and take over their computer, but I don't think I'd get away with that. A user will only hit this if they're screwing around trying to do things they shouldn't so I'm not too bothered about the experience at the mo.

src/ui/component/Learning.html
tav: Aha!

PR #122, opened 2018-03-02
Feature/462 update the rest of the lpg with the ui designs

jenbutongit:
PR includes:
* Reimplementing govuk scss. Mainly because they're using id selector to style the #content and they bypassed their own linter. (Just don't use # for css :S) Anyway, I'm using .container instead. It also helps when a section has a fullbleed background colour.
* New home layout [Tile component, resource item, bordered title]
* New learning record table which has a 'you have completed x out of y required learning'


src/ui/page/booking/confirm-booking.html
```diff
@@ -1,12 +1,15 @@
...
+        <div class="grid-row">
+            <div class="column-full">
+                <Breadcrumb :breadcrumbs/>
+                <h1 class="heading-large">
+                    <span class="heading-secondary">Booking a course</span>
+                    Confirm your details before completing your booking
+                </h1>
+            </div>
         </div>
+<<<<<<< HEAD
```
tav: The file needs to have conflicts resolved...

src/ui/page/home.html
```diff
@@ -1,56 +1,72 @@
...
-            {{else}}
-            <p>You have completed all your required learning.</p>
-            {{/if}}
-
-            <h2 class="heading-small">Learning plan</h2>
+                {{#if requiredLearning.length}}
+                <ul class="resource__list">
+                    {{#each requiredLearning as course}}
+                    <li class="resource">
+                        <Learning bind:course />
```
tav: I realise this was in a previous PR, but why are we doing `bind:course` here?

jenbutongit: I don't think I made this change, although I did notice when I was doing some of the Breadcrumb code, whenever I `<Breadcrumb :breadcrumbs/>` prettier would sometimes recompile to  `<Breadcrumb:breadcrumbs/>` and the server couldn't render the template. Maybe `bind:` bypasses it? unsure, doesn't make a difference though right? I can change it though

tav: Unless I've misunderstood `bind`, I think `course={{course}}` would be preferable?

src/ui/component/Learning.html
```diff
@@ -1,47 +1,45 @@
...
-            </ul>
-        </div>
+    </div>
+    <div class="resource__duration">
+        <span>{{course.type === ('elearning' || 'face-to-face') ? '' : course.type + ', '}}{{course.duration}}</span>
+    </div>
+    <div class="resource__status">
+        <span class="badge badge--info badge--smaller">{{i18n(course.state || 'not started')}}</span>
+        {{#if course.requiredBy && course.type !== ('elearning' || 'face-to-face') }}
+            <span class="meta">21 Aug 2018</span>
```
superafroman: What's this date meant to be? Date started, booked, progressed..? We'll need to add to the learner record.

src/ui/component/Learning.html
```diff
@@ -1,47 +1,45 @@
...
-                <li>{{i18n(course.type)}}</li>
-                <li>{{course.duration}}</li>
-            </ul>
-        </div>
+    </div>
+    <div class="resource__duration">
+        <span>{{course.type === ('elearning' || 'face-to-face') ? '' : course.type + ', '}}{{course.duration}}</span>
+    </div>
+    <div class="resource__status">
+        <span class="badge badge--info badge--smaller">{{i18n(course.state || 'not started')}}</span>
```
superafroman: I believe we should show no badge for 'not started'?

src/ui/component/Learning.html
jenbutongit: thanks for the catch, it's just course.requiredBy for non elearning/face to face

PR #123, opened 2018-03-02
Feature/learners can add suggestions to learning or remove suggestions

smoga-git:


package.json
```diff
@@ -37,6 +37,7 @@
 		"@types/pg": "7.4.3",
 		"@types/request": "2.47.0",
 		"@types/session-file-store": "1.2.0",
+		"body-parser": "^1.18.2",
```
tav: Did you mean to add this to `devDependencies` ?

tav: Also, explicit versions please, i.e. remove the `^`.

src/ui/controllers/suggestion.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export async function addToPlan(req: express.Request, res: express.Response) {
+	let uid = req.params.courseId
+	let verb = 'Liked'
+	const course = await catalog.get(uid)
+	const verbId = xapi.lookup(verb)
+
+	if (verbId && course) {
+		try {
+			let resp = await xapi.record(req, course, verbId)
```
tav: Do we care about the `resp` in this instance?

src/ui/controllers/suggestion.ts
```diff
@@ -0,0 +1,94 @@
...
+	res: express.Response
+) {
+	let uid = req.params.courseId
+	let verb = 'Disliked'
+	const course = await catalog.get(uid)
+	const verbId = xapi.lookup(verb)
+
+	if (verbId && course) {
+		try {
+			let resp = await xapi.record(req, course, verbId)
```
tav: Do we care about the `resp` in this instance?

src/lib/xapi.ts
```diff
@@ -123,14 +127,16 @@ export async function record(
 			},
 		}
 	}
-	await send(payload)
+	let resp = await send(payload)
+	return resp
```
tav: Can just `return await send(payload)`.

package.json
smoga-git: removed

src/lib/xapi.ts
smoga-git: deliberately returning the response thought it would be generally useful than just firing it off

smoga-git: Oh , right you are :D

PR #125, opened 2018-03-05
bug/453-bug-displaying-incorrect-date-selection-for-course-booking

superafroman:


src/lib/service/catalog/index.ts
```diff
@@ -466,9 +470,15 @@ export async function resetCourses() {
...
-	const esClient = await getElasticClient()
-	// delete index for reset
-	await esClient.indices.delete({index: 'dgraph'})
+
+	try {
+		const esClient = await getElasticClient()
+		// delete index for reset
+		await esClient.indices.delete({index: 'dgraph'})
+	} catch (e) {
+		console.error('Unable to delete elasticsearch index')
```
tav: For future, could we use the logger?

PR #129, opened 2018-03-05
feature/505-content-for-cancel-course-request-booking-page

jenbutongit:
PR includes:

* Course booked/cancelation confirmed copy (same as current CSL)
* Fix auto select first available date

### DIFF
--- a/src/ui/page/booking/cancel-booking.html
+++ b/src/ui/page/booking/cancel-booking.html
@@ -24,7 +24,7 @@
                         <span class="cancel__item--key">Requested Course:</span> <span class="bold-small">{{course.title}}</span>
                     </li>
                     <li>
-                        <span class="cancel__item--key">Requested Date:</span> <span class="bold-small">{{new Date(course.availability[0]).toDateString()}}</span>
+                        <span class="cancel__item--key">Requested Date:</span> <span class="bold-small">{{formatDate(selectedDate)}}</span>
                     </li>
                 </ul>
                 <form method="post" class="push-bottom">
--- a/src/ui/page/booking/cancel-booking.html
+++ b/src/ui/page/booking/cancel-booking.html
@@ -21,10 +21,10 @@
                 <h3 class="heading-small">{{signedInUser.givenName}}</h3>
                 <ul class="cancel__list push-bottom">
                     <li>
-                        <span class="cancel__item--key">Requested course:</span> <span class="bold-small">{{course.title}}</span>
+                        <span class="cancel__item--key">Requested Course:</span> <span class="bold-small">{{course.title}}</span>
                     </li>
                     <li>
-                        <span class="cancel__item--key">Requested date:</span> <span class="bold-small">{{new Date(course.availability[0]).toDateString()}}</span>
+                        <span class="cancel__item--key">Requested Date:</span> <span class="bold-small">{{new Date(course.availability[0]).toDateString()}}</span>
                     </li>
                 </ul>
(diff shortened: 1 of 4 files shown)


src/ui/page/booking/confirmed.html
```diff
@@ -3,7 +3,7 @@
         <div class="grid-row">
             <div class="column-full">
                 <div class="govuk-box-highlight">
-                    <h1 class="heading-xlarge">{{message}}</h1>
+                    <h1 class="heading-xlarge">{{message === 'Booked' ? 'Booking request submitted' : 'Bookign request cancelled' }}</h1>
```
superafroman: "Bookign"

src/ui/page/booking/choose-date.html
```diff
@@ -38,7 +38,7 @@ <h1 class="heading-large">
                                 <td>{{course.location}}</td>
                                 <td>&pound;{{course.price}}</td>
                                 <td>
-                                    {{#if availability.id == selectedDate}}
+                                    {{#if availability.id == (selectedDate + 1)}}
```
superafroman: `selectedDate` shouldn't be treated as a sequential index rather it's an ID - if there are dates that are too old to be shown then the first id won't be '0'. I intentionally defaulted the selectedDate to -1 in the controller so we don't select anything. As selecting the first means when a user clicks "Change date" their selection is reset to the first, better to not select anything than to select the wrong thing. 

If we want to select a date we need to sort going back to the page, otherwise we should leave it unselected.

src/ui/page/booking/confirmed.html
```diff
@@ -20,6 +21,10 @@ <h4 class="heading-small">What happens next?</h4>
                     Once confirmed you will be notified and we will contact you with the
                     course logistics.
                 </p>
+            {{else}}
+                <p>Your course booking has been cancelled.</p>
+                <p><a href="/profile">Visit your profile</a></p>
```
superafroman: Why would you go to your profile from this page? Wouldn't you want to go back to home?

src/ui/page/booking/cancel-booking.html
```diff
@@ -21,10 +21,10 @@ <h2 class="heading-medium error-summary-heading" id="error-summary-heading-examp
                 <h3 class="heading-small">{{signedInUser.givenName}}</h3>
                 <ul class="cancel__list push-bottom">
                     <li>
-                        <span class="cancel__item--key">Requested course:</span> <span class="bold-small">{{course.title}}</span>
+                        <span class="cancel__item--key">Requested Course:</span> <span class="bold-small">{{course.title}}</span>
                     </li>
                     <li>
-                        <span class="cancel__item--key">Requested date:</span> <span class="bold-small">{{new Date(course.availability[0]).toDateString()}}</span>
+                        <span class="cancel__item--key">Requested Date:</span> <span class="bold-small">{{new Date(course.availability[0]).toDateString()}}</span>
```
superafroman: Requested date isn't necessarily the first? The course record should identify what the selected date was.

src/ui/page/booking/confirmed.html
jenbutongit: this is from current csl 😧should we redirect to home?
screen shot 2018-03-05 at 15 30 35

superafroman: It's possible their profile is more like our home? I think we take them home - just like on booking request submitted there's a "You can now return to the home page" link.

PR #134, opened 2018-03-06
feature/436-match-use-of-xapi-events-to-the-video-profile-as-defined-by-the-xapi-video-cop-group

tav:


src/lib/xapi.ts
```diff
@@ -34,39 +48,70 @@ export interface Statement {
 	}
 }
 
+export enum Type {
+	Event = 'http://adlnet.gov/expapi/activities/event',
+	Face2Face = 'https://cslearning.gov.uk/types/course/face2face',
```
superafroman: FaceToFace 😛

src/lib/xapi.ts
```diff
@@ -106,27 +184,63 @@ export async function record(
...
+	// 		throw new Error(
+	// 			`Invalid value for the xAPI Progressed statement: "${valueJSON}"`
+	// 		)
+	// 	}
+	// 	payload.result = {
+	// 		extensions: {
+	// 			'https://w3id.org/xapi/cmi5/result/extensions/progress': value,
+	// 		},
+	// 	}
+	// }
```
superafroman: Are we holding onto this for a reason? Should remove commented code.

PR #135, opened 2018-03-06
chore/learner record auth

superafroman:


src/lib/xapi.ts
```diff
@@ -140,7 +141,7 @@ export async function record(
 			type = Type.Event
```
tav: Should this be `Type.Elearning` now?

src/ui/controllers/xapi.ts
```diff
@@ -12,7 +13,7 @@ export async function proxy(ireq: express.Request, res: express.Response) {
 
 	const agent = {
 		account: {
-			homePage: 'https://cslearning.gov.uk',
+			homePage: xapi.HomePage,
```
tav: Nice catch!

src/lib/xapi.ts
superafroman: Good spotting!

src/ui/controllers/xapi.ts
superafroman: Wouldn't have noticed but the missing trailing slash meant xapi didn't return results matching the user for these statements 🙄

PR #147, opened 2018-03-12
feature/485 spike allow multiple modules within a course

superafroman:
Updated domain model with new Module and Event classes. A Course now can contain
one or more Modules. If a Module is face-to-face it will have Events a User can
register for.

xAPI statements now reference the specific activity a user is interacting with
rather than just the course.

Moved to the new learning catalogue and removed dgraph. data.csv and
/courses/reset have been removed in favour of the new learning catalogue API
populating itself.


package.json
```diff
@@ -1,92 +1,85 @@
...
+    "@types/request": "2.47.0",
+    "@types/session-file-store": "1.2.0",
+    "@types/unzip": "0.1.0",
+    "@types/xml2js": "0.4.2",
+    "aws-sdk": "2.205.0",
+    "make-runnable": "1.3.6",
+    "node-sass": "4.7.2",
+    "nodemon": "1.14.12",
+    "protobufjs": "6.8.4",
+    "slugid": "^1.1.0",
```
tav: Fixed versions please...

src/bin/content.js
```diff
@@ -0,0 +1,226 @@
...
+    "HEO",
+    "SEO",
+    "G6",
+    "G7",
+    "SCS"
+];
+
+function parseCourses(file) {
+    const rawData = fs.readFileSync(file);
+    const lines = parse(rawData.toString());
```
tav: The canonical way of doing this is to specify `{encoding: 'utf8'}` as a parameter to the `fs.readFileSync()` call instead of having to call `.toString()` explicitly on the `Buffer` object.

src/lib/config/index.ts
```diff
@@ -141,11 +145,13 @@ export const SESSION_SECRET =
 	'dcOVe-ZW3ul77l23GiQSNbTJtMRio87G2yUOUAk_otcbL3uywfyLMZ9NBmDMuuOt'
 
 export const XAPI = set({
-	activityBaseUri: 'http://cslearning.gov.uk/courses',
 	auth: {
 		password: env.XAPI_PASS || '1c0e1b6827606d7efed71e204939d048f94f842b',
 		username: env.XAPI_USER || '66f2b4fc001e3da992d23b57d8a7457655bea078',
 	},
+	courseBaseUri: 'http://cslearning.gov.uk/courses',
```
tav: Should these URIs be singular, i.e. `/course` instead of `/courses`?

superafroman: I prefer plural on as requesting a /courses on an API would return a list, not a single course. Convention in xAPI is inconsistent so went with my preference :).

PR #152, opened 2018-03-14
Feature/537 update profile page design and allow a learner to edit their profile info

jenbutongit:
**PR includes:**

* New profile flow (view profile, click change, edit, confirmed), routes changed too
* Username is unchangable because of wso2, wso2 profession field (area of work) only allows string so using `.join()` and `.split()` to get around this
* ProfileEditPage.html component. Reusable, spits out one input per page, just give it the vars :) 
* Selects.html component, has some logic for Radios.html, Checkboxes.html, Typeahead.html. "Select" also has a defaulted 'other' field because _that's good form design™_ and also using 'other' in i18n is reserved for their plural thing 
* In Page.html add `{{{scripts}}}`, this is for typeahead component since we probably don't to render that js every page? 
* i18n'ed!


jenbutongit:
to note: I have been told that users will have a complete profile (?) ie, they wont go through a registration flow. Errors will still come up when a signed in user doesn't have a complete profile

jenbutongit:
[Update profile page design and allow a learner to edit their profile info](https://trello.com/c/YkyyDSyz/537-update-profile-page-design-and-allow-a-learner-to-edit-their-profile-info)

PR #155, opened 2018-03-14
Scorm blob upload   + Suggestion code change

smoga-git:


src/ui/controllers/suggestion.ts
```diff
@@ -67,13 +67,26 @@ export async function suggestionsForYou(
...
+) {
 	const suggestedLearning = (await catalog.findSuggestedLearning(user)).entries
-	const learningRecord = await learnerRecord.getLearningRecordOf(null, user)
+	let learningRecord: Record<string, model.Course> = {}
+	if (Object.keys(learningRecordIn).length > 0) {
+		learningRecord = learningRecordIn
+	} else {
+		learningRecord = hashArray(
+			await learnerRecord.getLearningRecordOf(null, user),
+			'uid'
```
superafroman: 'uid' is no longer on course - it's been updated to 'id'

PR #169, opened 2018-03-20
feature/485 allow multiple modules within a course

superafroman:
Updates so learning record supports modules. And fixed used of TERMINATED following on from video updates.


src/lib/xapi.ts
```diff
@@ -252,27 +254,28 @@ export async function record(
...
+				}
+				break
+			case 'link':
+				type = Type.Link
+				break
+			case 'video':
+				type = Type.Video
+				break
+			default:
+				throw new Error(`Unknown course type ${course.getType()}`)
```
tav: `module.type` ?

src/ui/component/LearningRecordTable.html
```diff
@@ -13,7 +13,7 @@
             <td><a href="/courses/{{course.id}}">{{course.title}} </a></td>
             <td>{{i18n(course.getType())}}</td>
             <td>{{course.getDuration() || '-'}}</td>
-            <td>{{formatDate(course.record.completionDate)}}</td>
+            <td>{{formatDate(course.getCompletionDate(signedInUser))}}</td>
```
tav: *throw up*

superafroman: Incremental improvements.

Dockerfile
```diff
@@ -39,8 +39,8 @@ ENV AUTHENTICATION_SERVICE_URL=http://identity.local.cshr.digital:9443 \
   PORT=3001 \
   SESSION_SECRET=topsecret
 
-# TODO(tav): Create an actual user so that we don't run this as root
```
tav: Leave the note in there until the root issue is addressed?

superafroman: Maybe it's cos I'm a cowboy, but I don't think this is such an issue when running inside containers.

PR #173, opened 2018-03-21
feature/564 update header design to reflect prototype

superafroman:


src/lib/ui/template.ts
```diff
@@ -161,6 +163,40 @@ function isFile(path: string) {
...
+				output += `<p>${line}</p>`
+			}
+		}
+		return output
+
+		//
+		// .filter(part => !!part)
+		// .map(part => `<p>${part}</p>`)
+		// .join()
+	}
```
m24murray: Do we need this comment?

src/ui/component/Header.html
```diff
@@ -2,8 +2,9 @@
     <div class="header-wrapper container">
         <div class="header-global">
             <div class="header-logo">
-                <a href="/" title="Go to the Learning Platform homepage" id="logo">
-                    Civil Service Learning
+                <a href="https://www.gov.uk" title="Go to the GOV.UK homepage" id="logo" class="content">
+                    <img src="img/gov.uk_logotype_crown_invert_trans.png" width="36" height="32" alt="">
```
m24murray: Do we need an alt tag?

src/lib/ui/template.ts
superafroman: Updated

src/ui/component/Header.html
superafroman: There isn't one in the prototype - I think it's OK here to omit cos it's just a logo.

PR #182, opened 2018-03-26
feature/517 added to learning plan notification replacement

superafroman:


superafroman:
@jenbutongit Gonna leave the message as per current story - so [Title] successfully added. [Title] has been successfully added to your learning plan

Once Rich has settled down his requirements we'll add new task and update.

PR #186, opened 2018-03-26
Feature/639 search pagination

m24murray:
**Overview:**
- Adding pagination to search results screen
- The latest version of Learning catalogue (with pagination changes will be required to test)

**Changes:**
- Pages and respective links should be dynamically added based upon search query, page size and page number
- The next/prev links should be hidden at the start and end of the page set
- The current page should not have a link
- If there are no results from a query, no result/page information should be shown
- If there is only one page of results, pagination links should not appear
- Adding _each range logic_ in templates courtesy of @tav 

**Concerns:**
- Currently the user can manually set the url to be e.g. page=99999, effectively setting the returning results to be out of bounds. Need to determine correct behaviour.
- If the user sets the page size to be 1, and there >10 results shown, it will display a page link for each of these page. This would be a problem if there are hundreds of results returned. However as the current data size is relatively small, we shouldn't have any major problems.

### DIFF
--- a/src/ui/page/search/index.html
+++ b/src/ui/page/search/index.html
@@ -87,6 +87,4 @@
         </div>
         {{/if}}
     </div>
-    <script>
-    </script>
 </Page>
--- a/src/ui/page/search/index.html
+++ b/src/ui/page/search/index.html
@@ -58,5 +58,50 @@
                 </div>
             </div>
         </div>
+        <div class="pager">
+            <div class="pager__controls">
+                <a class="pager__prev" href="#">Prev</a>
+                <ul class="pager__list">
+                    <li>
+                        <a href="#">
+                            1
+                        </a>
+                    </li>
+                    <li>
+                        2
+                    </li>
+                    <li>
+                        <a href="#">
+                            3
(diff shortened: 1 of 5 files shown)


src/ui/page/search/index.html
```diff
@@ -58,5 +58,35 @@ <h1 class="heading-xlarge heading heading--page-heading ">
...
+                {{#if searchResults.page != (Math.ceil(searchResults.totalResults/searchResults.size)-1)}}
+                <a class="pager__next" href="/search?q={{query}}&s={{searchResults.size}}&p={{searchResults.page + 1}}">Next</a>
+                {{/if}}
+            </div>
+            <div class="pager__summary">Showing {{((searchResults.page*searchResults.size)+1)}} – {{((searchResults.page*searchResults.size)+searchResults.results.length)}} of {{searchResults.totalResults}} items</div>
+        </div>
+        {{/if}}
     </div>
+    <script>
+    </script>
```
superafroman: Empty script block?

superafroman:
@m24murray needs a rebase

PR #199, opened 2018-03-28
Bug/659 remove security headers for test environment to enable automated testing

jenbutongit:


superafroman:
Tav added as well in IE fix branch.

PR #200, opened 2018-03-28
bug/661-update-there-were-1-results-for-start-to-there-was-1-result-for-start-ie-correct-pluralisation

jenbutongit:


src/ui/page/search/index.html
```diff
@@ -21,7 +21,7 @@ <h1 class="heading-xlarge heading heading--page-heading ">
                 {{/if}} {{#if searchResults && query}}
                 <p class="lede">There were
                     <strong class="lpg-search-amount">{{#if searchResults.totalResults}} {{searchResults.totalResults}} {{else }} no {{/if}}</strong>
-                    results for
+                    {{currentReq.__n('%s result', searchResults.totalResults)}} for
```
superafroman: Does this print out right? Looks to me (without running it) it'd print "There were 1 1 result for ..."

PR #206, opened 2018-03-29
update password

smoga-git:


tav:
Sorry @smoga-git, just realised that the same password validation logic needs to be on WSO2 as well as on UI due to using WSO2 for password reset 😢 — so we need to do the regex solution after all and might as well use the same regex in both places to make it easier to maintain.

PR #210, opened 2018-04-04
Added tests for search, updated user creds to fix tests, modified dis…

williampri:
…citecourse to classes

### DIFF
--- a/src/locale/en.json
+++ b/src/locale/en.json
@@ -31,8 +31,7 @@
 	"home_page_title": "Your learning",
 	"required_section_title": "Required learning",
 	"other_section_title": "Other learning",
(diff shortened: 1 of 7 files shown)


src/locale/en.json
```diff
@@ -256,5 +256,11 @@
 	"areasOfWork": "areasOfWork",
 	"audience_undefined_label": "audience_undefined_label",
```
jenbutongit: from here onwards, these key-values shouldn't be here

williampri: So... we need to move them? Or from now on?

jenbutongit: remove them

PR #216, opened 2018-04-05
bug/LPFG-99

smoga-git:
Removed requiredBy and frequency

### DIFF
--- a/src/management-ui/controllers/course/edit.ts
+++ b/src/management-ui/controllers/course/edit.ts
@@ -109,6 +109,9 @@ export async function setCourse(ireq: express.Request, res: express.Response) {
 
 		session.save(() => {
 			delete session.squash_session_clear
+			delete session.course
+			delete session.pendingFiles
+
 			res.redirect(`/courses`)
 		})
 	}


src/management-ui/controllers/course/edit.ts
```diff
@@ -142,7 +145,7 @@ export async function getCourse(ireq: express.Request, res: express.Response) {
 		await loadCourse(ireq, res)
 		course = req.course
 	}
-
+	console.log(course)
```
jenbutongit: remove console.log / replace with logger4js please

src/management-ui/controllers/course/edit.ts
```diff
@@ -142,7 +145,7 @@ export async function getCourse(ireq: express.Request, res: express.Response) {
 		await loadCourse(ireq, res)
 		course = req.course
 	}
-
+=
```
jenbutongit: is this = meant to be there?

smoga-git: doh!

PR #222, opened 2018-04-06
LPFG-111-Unable-to-re-add-removed-learning-to-the-learning-plan

smoga-git:
Add retrieved as a verb and check against this when adding course to planned learning


src/lib/xapi.ts
```diff
@@ -150,6 +150,7 @@ export const Verb = {
 	PlayedVideo: 'https://w3id.org/xapi/video/verbs/played',
 	Rated: 'http://id.tincanapi.com/verb/rated',
 	Registered: 'http://adlnet.gov/expapi/verbs/registered',
+	Retrieved: 'https://w3id.org/xapi/dod-isd/verbs/retrieved',
```
superafroman: What does retrieved mean?

src/ui/controllers/course/index.ts
```diff
@@ -180,6 +180,7 @@ export async function loadCourse(
 	const course = await catalog.get(courseId)
 	if (course) {
 		req.course = course
+		console.log(course)
```
superafroman: Remove or change to log4js with meaningful message.

src/ui/controllers/suggestion.ts
```diff
@@ -32,6 +32,14 @@ export async function addToPlan(ireq: express.Request, res: express.Response) {
 	}
 	try {
 		await xapi.record(req, course, xapi.Verb.Liked)
+
+		const user = req.user as model.User
+		const record = await learnerRecord.getRecord(user, course) // check record for archived state
+		// need someway to indicate 'unarchived' state retrieved seems to be closest verb
+		if ((record as learnerRecord.CourseRecord).state === 'ARCHIVED')
+			console.log('seting unarchived')
```
superafroman: We shouldn't need to do this - setting 'Liked' should have the same effect as setting unarchived. If this isn't the case then learner-record should be updated so that it does.

src/ui/controllers/course/index.ts
jenbutongit: please remove or use logger4js

src/lib/xapi.ts
smoga-git: "Get or bring something back; regain possession of.* ;D

If an article is removed it is archived which changes it's state. Adding or removing to learning plan changes its preference. There doesn't seem to be a way of removing a state once changed? so we need a way of saying something iks no longer archived when it is readded once it is removed.

src/ui/controllers/course/index.ts
smoga-git: Removed

smoga-git: Removed

PR #223, opened 2018-04-09
LPFG-97-Cost-missing-from-classroom-base-course

smoga-git:
Course price calculated from module
Added price field to face to face module form

### DIFF
--- a/src/lib/config/index.ts
+++ b/src/lib/config/index.ts
@@ -48,7 +48,7 @@ export const AUTHENTICATION = set({
 	servicePassword: 'changeme!12',
 	serviceUrl:
 		env.AUTHENTICATION_SERVICE_URL ||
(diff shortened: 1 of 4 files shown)


src/lib/config/index.ts
```diff
@@ -58,11 +58,8 @@ export const BOOKING_CONFIRMED_NOTIFY_TEMPLATE_ID =
 	'8efb52bd-9ada-402e-8fab-84a751bf4a71'
 
 export const BOOKING_NOTIFY_RECIPIENTS = [
-	// 'david.acquet@cabinetoffice.gov.uk',
```
williampri: Have these people decided not to be part of the gov notify project now?

smoga-git: Not removed by this branch

PR #224, opened 2018-04-09
[DO NOT MERGE]feature/Lpfg-111 unable to re add removed learning to the learning plan

smoga-git:


src/ui/controllers/course/index.ts
```diff
@@ -180,6 +180,7 @@ export async function loadCourse(
 	const course = await catalog.get(courseId)
 	if (course) {
 		req.course = course
+		console.log(course)
```
superafroman: Remove or use log4js with meaningful message :)

src/lib/xapi.ts
```diff
@@ -150,6 +150,7 @@ export const Verb = {
 	PlayedVideo: 'https://w3id.org/xapi/video/verbs/played',
 	Rated: 'http://id.tincanapi.com/verb/rated',
 	Registered: 'http://adlnet.gov/expapi/verbs/registered',
+	Retrieved: 'https://w3id.org/xapi/dod-isd/verbs/retrieved',
```
superafroman: As we're no longer using it lets get rid of `Retrieved` from here, below and in the template.

PR #240, opened 2018-04-23
test/lpfg-155

jenbutongit:
**PR includes**

* One user is logged in, and there is a pause to allow them time to log in, and their cookie is stored globally in `artilleryfunc.js` and this is used to set the cookie for subsequent users

### DIFF
--- a/test/performance/src/artilleryfunc.js
+++ b/test/performance/src/artilleryfunc.js
@@ -17,7 +17,7 @@ let triedLogin = false
 
 function useCookie(requestParams, context, ee, next) {
 	if (Object.keys(testCookie).length !== 0) {
-		requestParams.headers.cookie = testCookie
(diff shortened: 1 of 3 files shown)


test/performance/src/artilleryfunc.js
```diff
@@ -1,11 +1,70 @@
...
 	sessionDataKey: sessionDataKey,
+	loginToWso2: loginToWso2,
+	logBeforeRequest: logBeforeRequest,
+	useCookie: useCookie,
+}
+
+let testCookie = {}
+let triedLogin = false
+
+function useCookie(requestParams, context, ee, next) {
```
superafroman: Is this ever used? There's a bug - `requestParams.headers.cookie = {'lpg-ui': testCookie}` should be `requestParams.headers.cookie = testCookie`. But if the function isn't used just remove.

PR #241, opened 2018-04-24
bug/LPFG-190

jenbutongit:
**PR includes**

* styling fix for DisciteCourse - progress badge was "pushing down" the other course info, so it would overflow if the course description was too short
* removed margin-bottom for the module counter thingy
* Fix for video times. duration was not being saved to module, so there would be either no time, or an old time

### DIFF
--- a/src/management-ui/controllers/module/edit.ts
+++ b/src/management-ui/controllers/module/edit.ts
@@ -208,7 +208,8 @@ export async function setModule(ireq: express.Request, res: express.Response) {
 				res.sendStatus(500)
 				return
 			}
-			req.course.duration = duration
+			// req.course.duration = duration
+			data.duration = duration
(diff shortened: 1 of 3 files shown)


src/management-ui/controllers/module/edit.ts
```diff
@@ -208,7 +208,8 @@ export async function setModule(ireq: express.Request, res: express.Response) {
 				res.sendStatus(500)
 				return
 			}
-			req.course.duration = duration
+			// req.course.duration = duration
```
superafroman: Remove comment

superafroman: 😛

jenbutongit: removed 😛

PR #249, opened 2018-05-01
LPFG-145 modules in search results

smoga-git:


src/ui/component/DisciteModule.html
```diff
@@ -1,24 +1,43 @@
 <div class="discite__item u-clearfix discite__item--module">
-    <div class="discite__main discite__main--module">
-        <h3 class="heading bold-small heading--text">
-            {{module.title}}
-        </h3>
+    <div class="discite__main discite__main--module--{{modifier || ''}}">
```
jenbutongit: in BEM there shouldn't be any cases where you have a modifier modifier. I've created an element  .discite__counter. Could you put this in there? refer to `DisciteCourse.html` for structure.

src/ui/assets/styles/components/_discite.scss
```diff
@@ -38,6 +38,12 @@ $module: 'discite';
 		@include media(desktop) {
 			padding-right: 300px;
 		}
+
+		@include m(module) {
```
jenbutongit: If you need to you can use discite__counter and add the modifier `@include m(module)` in there, and get rid of the nested modifier

src/ui/component/DisciteModule.html
smoga-git: done

src/ui/assets/styles/components/_discite.scss
smoga-git: done

PR #251, opened 2018-05-02
LPFG-231 - Course completion link still using old url

m24murray:
https://cshrdigitalandanalysis.atlassian.net/browse/LPFG-231

- fixing upload url
- removing old format of urls
- linting


m24murray:
Verified that a new course uploaded contains the correct host format in both [player_overrides](url) and `close_methods`

PR #268, opened 2018-05-10
bug/LPFG-235 express-validator

jenbutongit:
checks every field in req.body and escapes

### DIFF
--- a/src/ui/server.ts
+++ b/src/ui/server.ts
@@ -28,9 +28,9 @@ import * as userController from './controllers/user'
 import * as xApiController from './controllers/xapi'
 
 /* tslint:disable:no-var-requires */
-const favicon = require('serve-favicon')
-/* tslint:disable:no-var-requires */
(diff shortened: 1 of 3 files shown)


src/ui/server.ts
```diff
@@ -27,10 +27,10 @@ import * as suggestionController from './controllers/suggestion'
 import * as userController from './controllers/user'
 import * as xApiController from './controllers/xapi'
 
-/* tslint:disable:no-var-requires */
-const favicon = require('serve-favicon')
 /* tslint:disable:no-var-requires */
 const flash = require('connect-flash')
+const favicon = require('serve-favicon')
+/* tslint:enable */
```
elxsy: INFO - What happened to tslinting?

superafroman: Certain modules haven't been packaged to support `import` syntax so need to be required, which tslint rejects.

PR #269, opened 2018-05-10
Feature/lpfg-288

jenbutongit:
**PR includes**
* helper structs 
* get levels from registry service

**To do**
* implement endOfBranch logic
* update user

### DIFF
--- a/src/lib/config/index.ts
+++ b/src/lib/config/index.ts
@@ -118,7 +118,7 @@ export const LOGGING = set(
 				out: {type: 'console'},
 			},
 			categories: {
-				default: {appenders: ['out'], level: 'all'},
+				default: {appenders: ['out'], level: 'debug'},
 			},
 		},
 	}
@@ -142,3 +142,5 @@ export const XAPI = set({
 })
 
 export const YOUTUBE_API_KEY = env.YOUTUBE_API_KEY
+
+export const registryServiceURL = 'http://localhost:9002'
--- a/src/lib/config/index.ts
+++ b/src/lib/config/index.ts
@@ -143,4 +143,5 @@ export const XAPI = set({
 
 export const YOUTUBE_API_KEY = env.YOUTUBE_API_KEY
 
-export const registryServiceURL = 'http://localhost:9002'
+export const REGISTRY_SERVICE_URL =
+	env.REGISTRY_SERVICEURL || 'http://localhost:9002'
--- a/src/ui/component/Checkboxes.html
+++ b/src/ui/component/Checkboxes.html
@@ -1,5 +1,5 @@
 <fieldset class="u-space-b30">
-    {{#each options as option}}
+    {{#each options as option, index}}
         <div class="multiple-choice">
             <input id="{{option[0]}}" type="checkbox" name="{{inputName}}" value="{{option[0]}}" checked="{{(value || []).indexOf(option[0]) > -1 ? 'checked' : ''}}">
             <label for="{{option[0]}}">{{option[1]}}</label>
--- a/src/ui/component/Checkboxes.html
+++ b/src/ui/component/Checkboxes.html
@@ -1,5 +1,5 @@
 <fieldset class="u-space-b30">
-    {{#each options as option, index}}
+    {{#each options as option}}
         <div class="multiple-choice">
             <input id="{{option[0]}}" type="checkbox" name="{{inputName}}" value="{{option[0]}}" checked="{{(value || []).indexOf(option[0]) > -1 ? 'checked' : ''}}">
             <label for="{{option[0]}}">{{option[1]}}</label>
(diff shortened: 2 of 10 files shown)


src/lib/config/index.ts
```diff
@@ -142,3 +142,5 @@ export const XAPI = set({
 })
 
 export const YOUTUBE_API_KEY = env.YOUTUBE_API_KEY
+
+export const registryServiceURL = 'http://localhost:9002'
```
superafroman: Allow env override. Also think our convention is all caps?

src/ui/component/Checkboxes.html
```diff
@@ -1,5 +1,5 @@
 <fieldset class="u-space-b30">
-    {{#each options as option}}
+    {{#each options as option, index}}
```
superafroman: Are we using `index`?

src/ui/controllers/user.ts
```diff
@@ -127,57 +134,146 @@ export enum OptionTypes {
...
+		req.session!.levels = []
+		levels = []
+		//if there are no levels selected
+		const traversonResult = await traverson
+			.from('http://localhost:9002')
+			.jsonHal()
+			.follow('professions')
+			.getResource().result
+		try {
+			const parsed = parseRegistryProfiles(traversonResult)
```
superafroman: Can we just use the result obj?

src/ui/controllers/user.ts
```diff
@@ -127,57 +134,146 @@ export enum OptionTypes {
...
+				} else {
+					followPath.push(`jobRoles[${selection}]`, 'children')
+				}
+			})
+		}
+
+		const traversonResult = await traverson
+			.from('http://localhost:9002')
+			.jsonHal()
+			.follow(followPath)
```
superafroman: This could be more efficient if we start at the self link of the previous selection. Each link in follow makes another request to the server.

src/ui/controllers/user.ts
```diff
@@ -127,57 +134,146 @@ export enum OptionTypes {
...
+			.getResource().result
+		try {
+			const parsed = parseRegistryRoles(traversonResult)
```
superafroman: Latest registry has the `hasChildren` flag now :)

src/ui/component/Checkboxes.html
jenbutongit: oops not meant to be in this PR

src/ui/controllers/user.ts
```diff
@@ -127,57 +134,149 @@ export enum OptionTypes {
...
+	} else {
+		/* if the user has selected levels or there are req.params[0] */
+		levels = req.session!.levels
+		const followPath: string[] = []
+		prevLevelUrl = req.session!.prevLevelUrl
+
+		if (levels.length === selectedArr.length - 1) {
+			/* check the session. If the amount of levels saved matches up with the amount of selections made */
+			selectedArr.length === 1
+				? followPath.push(`professions[${selected}]`, 'jobRoles')
```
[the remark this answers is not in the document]
jenbutongit: Now using index 'input' which maps to a profession/jobRole which is stored in session, and it's url is read to get data

jenbutongit:
branch [LPG-130-Retrieve-and-Update-Civil-Servant-Details](https://github.com/Civil-Service-Human-Resources/lpg-services/tree/LPG-130-Retrieve-and-Update-Civil-Servant-Details) has now rebased onto this and any further changes will be done there.  👋 

**The changes made in this branch include**
* Querying Registry Service for professions and job roles
* Helper functions to parse the traverson data
* Logic to show the levels 
* On the profession/job role levels selection page, when selecting a user it will append to the url `/${selection}`. `${selection}` maps directly to an object from the response
  * The object has three properties `self`, `name` (and `hasChildren` if it is a jobRole)
  * The `self` is a url to this object on the Registry Service and it is used to query the next call, using `.from(${self}).follow(${children})`. 
  * If `hasChildren` is false, there is no call to it's children, and instead on the selection page, a query is appended to the url. `select=${profession || jobRole}/${object.id}`. This is used to update the user.

jenbutongit:
branch [LPG-130-Retrieve-and-Update-Civil-Servant-Details](https://github.com/Civil-Service-Human-Resources/lpg-services/tree/LPG-130-Retrieve-and-Update-Civil-Servant-Details) has now rebased onto this and any further changes will be done there.  👋 

**The changes made in this branch include**
* Querying Registry Service for professions and job roles
* Helper functions to parse the traverson data
* Logic to show the levels 
* On the profession/job role levels selection page, when selecting a user it will append to the url `/${selection}`. `${selection}` maps directly to an object from the response
  * The object has three properties `self`, `name` (and `hasChildren` if it is a jobRole)
  * The `self` is a url to this object on the Registry Service and it is used to query the next call, using `.from(${self}).follow(${children})`. 
  * If `hasChildren` is false, there is no call to it's children, and instead on the selection page, a query is appended to the url. `select=${profession || jobRole}/${object.id}`. This is used to update the user.

PR #276, opened 2018-05-14
LPFG-332 - Adding option to remove learning from optional plan

m24murray:
- Fixing issue where Other Learning had "Add to learning plan", and not have an link to remove learning from plan. 
- Adding flash message on home page when course is removed


superafroman:
@m24murray can you rebase?

PR #278, opened 2018-05-15
Feature/lpfg 130 retrieve and update civil servant details

smoga-git:
Okay, this pulls in details from identity service and registry to create a user profile. User profile can be edited and the changes saved to the registry

**Jen's comments on branch**

Comment on #269 Feature/lpfg-288
branch LPG-130-Retrieve-and-Update-Civil-Servant-Details has now rebased onto this and any further changes will be done there. :wave:

*The changes made in this branch include*

• Querying Registry Service for professions and job roles
• Helper functions to parse the traverson data
• Logic to show the levels
• On the profession/job role levels selection page, when selecting a user it will append to the url `/${selection}`. `${selection}` maps directly to an object from the response
   • The object has three properties `self`, `name` (and `hasChildren` if it is a jobRole)
   • The `self` is a url to this object on the Registry Service and it is used to query the next call, using `.from(${self}).follow(${children})`.
   • If `hasChildren` is false, there is no call to it's children, and instead on the selection page, a query is appended to the url. `select=${profession || jobRole}/${object.id}`. This is used to update the user.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
 		"@types/node": "9.4.0",
 		"@types/node-fetch": "1.6.7",
 		"@types/passport": "0.4.2",
-		"@types/passport-saml": "0.15.5",
 		"@types/request": "2.47.0",
 		"@types/session-file-store": "1.2.0",
 		"@types/shortid": "0.0.29",
@@ -48,15 +47,16 @@
 		"@types/xml2js": "0.4.2",
 		"csv": "2.0.0",
 		"make-runnable": "1.3.6",
-		"node-sass": "4.7.2",
 		"nodemon": "1.14.12",
 		"protobufjs": "6.8.4",
 		"slugid": "1.1.0",
+		"traverson-hal": "6.0.0",
 		"tslint": "5.9.1",
 		"typescript": "2.7.1"
 	},
 	"dependencies": {
 		"@types/express-validator": "3.0.0",
+		"@types/traverson": "2.0.28",
 		"axios": "0.17.1",
 		"azure-storage": "2.8.1",
 		"compression": "1.7.1",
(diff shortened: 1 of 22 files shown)


package.json
```diff
@@ -48,15 +47,16 @@
...
 		"nodemon": "1.14.12",
 		"protobufjs": "6.8.4",
 		"slugid": "1.1.0",
+		"traverson-hal": "^6.0.0",
 		"tslint": "5.9.1",
 		"typescript": "2.7.1"
 	},
 	"dependencies": {
 		"@types/express-validator": "3.0.0",
+		"@types/traverson": "^2.0.28",
```
jenbutongit: Please use exact versions

package.json
```diff
@@ -91,6 +93,8 @@
 		"striptags": "3.1.1",
 		"svelte": "1.55.0",
 		"tmp": "0.0.33",
+		"traverson-hal": "^6.0.0",
+		"traverson-promise": "0.0.9",
```
jenbutongit: If you decided not to use traverson-promise please remove

src/lib/config/index.ts
```diff
@@ -142,3 +142,6 @@ export const XAPI = set({
 })
 
 export const YOUTUBE_API_KEY = env.YOUTUBE_API_KEY
+
+export const REGISTRY_SERVICE_URL =
+	env.REGISTRY_SERVICEURL || 'http://localhost:9002'
```
jenbutongit: REGISTRY_SERVICE_URL?

superafroman: Missing an underscore :)

src/lib/config/passport.ts
```diff
@@ -104,8 +108,8 @@ export function hasRole(role: string) {
...
 
 export function logout(req: express.Request, res: express.Response) {
-	strategy.logout(req, (err, url) => {
-		req.logout()
-		res.redirect(url)
-	})
+	// strategy.logout(req, (err, url) => {
+	// 	req.logout()
+	// 	res.redirect(url)
+	// })
```
superafroman: What's the deal here? No longer supported? If there's no need to do a logout on the strategy can we still do a `req.logout()` and `res.redirect(url)`?

PR #279, opened 2018-05-15
Feature/LPFG-13 - Adding cookie page and banner

m24murray:
- adding cookies page and banner (cookie page has fairly basic styling for now, but may need stylised in line with gds in future)
- adding cookie controller to set cookie seen_cookie_message value to true
- js to check value of cookie and hide banner accordingly
- if user's browser has javascript turned off, the message should always appear

### DIFF
--- a/src/ui/component/Page.html
+++ b/src/ui/component/Page.html
@@ -15,6 +15,12 @@
     <script async src="https://www.googletagmanager.com/gtag/js?id={($config.GOOGLE_ANALYTICS_ID)}"></script> {@html $config.GOOGLE_ANALYTICS_CODE}
 </head>
 <body>
+    <div id="global-cookie-message">
+        <p>
+            GOV.UK uses cookies to make the site simpler.
+            <a href="/cookies"> Find out more about cookies</a>
+        </p>
+    </div>
     <Header />
     <Layout {_csrf}>
         <slot> </slot>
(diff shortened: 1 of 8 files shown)


src/ui/component/Page.html
```diff
@@ -14,6 +14,12 @@
     <script async src="https://www.googletagmanager.com/gtag/js?id={{config.GOOGLE_ANALYTICS_ID}}"></script> {{{config.GOOGLE_ANALYTICS_CODE}}}
 </head>
 <body>
+    <div id="global-cookie-message">
+        <p>
+            GOV.UK uses cookies to make the site simpler.
```
superafroman: Probably need to review this message - can you run past Rich?

src/ui/server.ts
```diff
@@ -122,7 +141,8 @@ app.use(
 app.use(serveStatic('assets'))
 app.use(favicon(path.join('assets', 'img', 'favicon.ico')))
 
-passport.configure('lpg-ui', config.AUTHENTICATION.serviceUrl, app)
+passport.configure('lpg-ui', config.AUTHENTICATION.	serviceUrl, app)
```
superafroman: tab/space before `serviceUrl`?

src/ui/component/Page.html
m24murray: Changing to Civil Service Learning

m24murray:
This should be ready to go now, if this gets out of date and i'm not around can someone else rebase and push? 🙏🛠👉

PR #281, opened 2018-05-16
hook-up-admin-and-identity

m24murray:
- adding missing traverson dependancy
- adding passport config for management
- links to identity service PR - https://github.com/Civil-Service-Human-Resources/identity-service/compare/fix/LPFG-344-hook-up-admin-and-identity

### DIFF
--- a/src/management-ui/server.ts
+++ b/src/management-ui/server.ts
@@ -69,11 +69,17 @@ app.use(compression({threshold: 0}))
 app.use(serveStatic('assets'))
 app.use(favicon(path.join('assets', 'img', 'favicon.ico')))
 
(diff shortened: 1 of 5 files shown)


src/management-ui/server.ts
```diff
@@ -69,11 +69,17 @@ app.use(compression({threshold: 0}))
...
+	'test',
+	config.AUTHENTICATION.serviceUrl,
+	app,
+	config.LPG_MANAGMENT_SERVER
+	)
 i18n.configure(app)
 
 app.use(passport.isAuthenticated)
-app.use(passport.hasRole('management'))
+app.use(passport.hasRole('USER'))
```
m24murray: setting to USER for now but we may need to change the config of management roles on the identity service

PR #285, opened 2018-05-21
feature/LPFG-338

jenbutongit:
**PR includes**

Generally, this is mostly an extension of the elearning uploading. elearning uploads have not changed just sidestepped using if statements.
* Added file support    
* file validator
  * input only supports .xls,.xlsx,.doc,.docx,.ppt,.pptx,.pdf,.ppsm types however drag+drop for any file type is not prevented
  * when a user tries to upload a file the validator in the client server will throw errors based on file meta data and file extension.
   It will redirect the user back to the page with a flash message. (They currently never upload directly to the CDN)
* lpg-catalogue-service and lpg-services have been updated so Module object contains fileSize and url property. This docker file is also pointing to the correct learning-catalogue-service tag
* file helpers are included via template.ts. 
  * fileName - parses url for the last fragment, and optionally accepts withExtension, which is a boolean. (Gets the file name, with extension?)
  * extension - gets the extension only from a filename or url.
  * appropriateFileSize - fileSize is saved in bytes. Depending on how many bytes the most appropriate unit will be used (B, KB, MB, etc)
* DisciteModule has been updated for files, since they look slightly different from other modules. (Another if statement.. sorry! :-( )

### DIFF
--- a/src/lib/config/passport.ts
+++ b/src/lib/config/passport.ts
@@ -30,7 +30,7 @@ export function configure(
 			profile: any,
 			cb: oauth2.VerifyCallback
 		) => {
-			// profile.accessToken = accessToken
+			profile.accessToken = accessToken
 			// get details here
 			const identityDetails = await identity.getDetails(accessToken)
 			const regDetails = await registry.profile(accessToken)
--- a/src/lib/config/passport.ts
+++ b/src/lib/config/passport.ts
@@ -33,23 +33,18 @@ export function configure(
 			profile: any,
 			cb: oauth2.VerifyCallback
 		) => {
-			profile.accessToken = accessToken
+			// profile.accessToken = accessToken
+			// get details here
+			const identityDetails = await identity.getDetails(accessToken)
+			const regDetails = await registry.profile(accessToken)
 
-			try {
-				const identityDetails = await identity.getDetails(accessToken)
-				const regDetails = await registry.profile(accessToken)
-
-				const combined = {
-					...profile,
-					...identityDetails,
-					...regDetails,
-				}
-				const user = model.User.create(combined)
-				return cb(null, user)
-			} catch (e) {
-				logger.warn(`Error retrieving user profile information`, e)
-				cb(e)
+			const combined = {
+				...profile,
+				...identityDetails,
+				...regDetails,
 			}
+			const user = model.User.create(combined)
+			return cb(null, user)
 		}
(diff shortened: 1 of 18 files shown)


src/lib/config/passport.ts
```diff
@@ -33,23 +30,18 @@ export function configure(
 			profile: any,
 			cb: oauth2.VerifyCallback
 		) => {
-			profile.accessToken = accessToken
-
-			try {
-				const identityDetails = await identity.getDetails(accessToken)
-				const regDetails = await registry.profile(accessToken)
+			// profile.accessToken = accessToken
```
superafroman: Don't we need this for further request to e.g. registry-service? If not delete the line rather than commenting.

src/lib/filehelpers.ts
```diff
@@ -0,0 +1,19 @@
+export function appropriateFileSize(fileSize: number): string {
+	const c = 1024
+	const dp = 2 //number of decimal places
+	const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
```
superafroman: Support for Yobibyte already - nice... 😱

src/ui/controllers/course/index.ts
```diff
@@ -99,8 +99,9 @@ export async function displayModule(
 			res.redirect(`/book/${course.id}/${module.id}/choose-date`)
 			break
 		case 'link':
+		case 'file':
 			await xapi.record(req, course, xapi.Verb.Experienced, undefined, module)
-			res.redirect(module.location!)
+			res.redirect(module.location! || module.url!)
```
superafroman: This looks wrong - but I've been doing some swift so might be getting confused. Force unwrapping module.location will throw an error before the or has a chance to check?

src/lib/filehelpers.ts
jenbutongit: just in case 😃 #scalable

src/ui/controllers/course/index.ts
jenbutongit: Swift is what typescript wishes it was. Anyway, .location is "optional", so unwrap and check if its there, if it's undefined, try module.url

jenbutongit: although, if both fails, who knows what'll happen 😛

PR #288, opened 2018-05-22
hotfix/using management id

jenbutongit:


src/management-ui/server.ts
```diff
@@ -72,7 +72,7 @@ app.use(serveStatic('assets'))
 app.use(favicon(path.join('assets', 'img', 'favicon.ico')))
 
 passport.configure(
-	config.AUTHENTICATION.clientId,
+	config.AUTHENTICATION.managementId,
```
superafroman: Need to do the same for secret!

PR #309, opened 2018-06-01
Lpfg 390  link up front end gui to backend

smoga-git:


superafroman:
I'm assuming this failing build is due to identity and civil servant registry not being up-to-date on dev2. Will restart it once I've updated the environment.

PR #310, opened 2018-06-01
Hotfix/profile-prg

jenbutongit:


src/ui/controllers/user.ts
```diff
@@ -507,31 +508,28 @@ export async function updateProfile(
 			break
 		case 'lineManager':
 			if (fieldValue !== req.body.confirmLineManager) {
-				lineManagerFailed = 'Line Manager Email does not match the confirmation.'
+				errorMessage = 'Line Manager Email does not match the confirmation.'
 			} else if (!validEmail.exec(fieldValue)) {
-				lineManagerFailed = 'Line Manager Email  is not valid'
+				errorMessage = 'Line Manager Email  is not valid'
```
superafroman: An extra space in this message after "Email". Also, could we extract all these messages out into the i18n file?

PR #335, opened 2018-06-14
Feature/lpfg 369 lpfg 368

jenbutongit:
**PR includes**
- A separate template had be made for confirmation emails for line manager and for learners
- config vars changed to group the template ids more nicely
- [x] [LPFG-368](https://cshrdigitalandanalysis.atlassian.net/browse/LPFG-368)
- [x] [LPFG-369](https://cshrdigitalandanalysis.atlassian.net/browse/LPFG-369)
- [x] needs squashing after #332 is merged

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -71,8 +71,8 @@ services:
     container_name: civil-servant-registry
     environment:
       - SPRING_PROFILES_ACTIVE=default
-      - CHECK_TOKEN_URL=http://identity:8080/oauth/check_token
-    image: cshr/civil-servant-registry-service:9fe7cd23d21b7428786e14f6a8c3a548b6a7211f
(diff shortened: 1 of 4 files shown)


docker-compose.yaml
```diff
@@ -71,8 +71,8 @@ services:
     container_name: civil-servant-registry
     environment:
       - SPRING_PROFILES_ACTIVE=default
-      - CHECK_TOKEN_URL=http://identity:8080/oauth/check_token
-    image: cshr/civil-servant-registry-service:9fe7cd23d21b7428786e14f6a8c3a548b6a7211f
```
smoga-git: Should this be changed?

superafroman: Yeah, check token url is still used but derived from the oauth service url now along with a couple of other endpoints.

PR #336, opened 2018-06-14
Feature/lpfg-317 learning record is updated after a booking

smoga-git:


src/ui/component/DisciteCourse.html
```diff
@@ -30,7 +30,7 @@
     </div>
 
     <slot name="actions">
-        <DisciteCourseActions {course} {modifier} {module} />
+        <DisciteCourseActions {course} {modifier} {module} {today}/>
```
jenbutongit: You can just do `today={new Date()}` and then remove it everywhere else. It stops you from repeating since you're only using it one place, then you don't have to continually keep passing `today` down to the page -> component -> component.

Alternatively you can add it to datetime.ts and within DisciteCourseActions call it with $datetime if you think we are going to use it more often

src/ui/page/home.html
```diff
@@ -52,7 +55,7 @@ <h2 class="heading-medium heading heading--red-top lpg-other-learning">
                 {#if plannedLearning.length}
                 <div class="resources lpg-other-courses">
                     {#each plannedLearning as course}
-                    <DisciteCourse course={course} modifier="home" /> {/each}
+                    <DisciteCourse course={course} modifier="home" today="{today}" /> {/each}
```
jenbutongit: See the other comment about `{today}`. Also with svelte v2 you can just do {today}, or today={today} (without the double `"`)

PR #337, opened 2018-06-15
chore/updating docker-compose.yaml

m24murray:
- updating docker tags
- getting rid of redundant env vars (these are defaulted in service anyway)


m24murray:
I'm going to close this PR and update with a newer version of this file

PR #339, opened 2018-06-18
feature/LPFG-372 LPFG-373

jenbutongit:
**this PR includes:**

* added cancelled line manager gov notify booking ID
* booking controller was getting long. split into booking and cancel


src/locale/en.json
```diff
@@ -231,7 +231,7 @@
 	"action_IN_PROGRESS": "Resume",
 	"action_NOT_STARTED": "Start",
 	"action_video_COMPLETED": "Restart",
-	"action_REGISTERED": "Requested",
+	"action_REGISTERED": "action_REGISTERED",
```
smoga-git: Is this correct?

src/ui/controllers/booking/booking.ts
```diff
@@ -5,84 +5,16 @@ import * as learnerRecord from 'lib/learnerrecord'
...
-) {
-	const req = ireq as extended.CourseRequest
-	const course = req.course
-	const module = req.module!
-	const event = req.event!
-
-	const record = await learnerRecord.getRecord(req.user, course, module, event)
-
-	if (!record) {
-		logger.warn(
```
smoga-git: This is repeated three times, better to abstract to a function rather than repeat the same thing three times. If you rebase you'll find this has been done in booking though!

src/ui/controllers/booking/cancel.ts
```diff
@@ -0,0 +1,154 @@
...
+	const course = req.course
+	const module = req.module!
+	const event = req.event!
+
+	const record = await learnerRecord.getRecord(req.user, course, module, event)
+
+	if (!record) {
+		logger.warn(
+			`Attempt to cancel a booking when not registered. user: ${
+				req.user.id
```
smoga-git: Same message as in booking , maybe make this an export somewhere?

src/locale/en.json
jenbutongit: oops!

PR #340, opened 2018-06-18
Lpfg 494  update lpg services admin ui to indicate relevant interests AND Update UI to show suggested learning based on interests

smoga-git:


src/management-ui/component/Audience.html
```diff
@@ -46,6 +46,16 @@ <h2 class="heading heading-medium ">
                 <button name="submit" type="submit" class="button audience__button" value="change_{i}_grade">Change</button>
             </td>
         </tr>
+        <tr>
+            <td class="key-value-list__key key-value-list__key--auto">{$i18n('audience_interest_label')}</td>
+            <td class="key-value-list__value key-value-value--auto lpg-interest">
+                {#each audience.interests as interest}
```
jenbutongit: Are we pulling interests from CSRS? Maybe we should just output the values from that rather than going through i18n otherwise we might end up with mismatches if the en.json isn't kept up to date. Can open a new jira ticket for this though if this is working fine

PR #343, opened 2018-06-19
LPFG 421 Setting Audience removes uploaded file

smoga-git:


src/management-ui/controllers/module/edit.ts
```diff
@@ -192,6 +192,7 @@ export async function setModule(ireq: express.Request, res: express.Response) {
 	const session = req.session!
 
 	req.body.location = req.sanitizeBody('location').unescape()
+	req.body.url = req.sanitizeBody('url').unescape()
```
superafroman: Possibly worth removing these hidden inputs and updating teh following line to something like this - 
```
const data = {
  ...module!,
  ...req.body,
  events: module!.events || [],
  id: module!.id || 'add-module', //temp id for persistance
  type: module!.type || req.params.moduleType,
}
```
This results in any data not overridden by the request body being retained. Haven't reviewed the whole process so not 100% sure it fits all cases but if it does would prevent us having to add further hidden inputs as we add more data to modules.

smoga-git: fair play

PR #346, opened 2018-06-21
Lpfg 355 optional modules within a course

smoga-git:


m24murray:
on hold while regression bug with mandatory checkbox is resolved

PR #347, opened 2018-06-21
chore/call-to-actions-LPFG-426-LPFG-302

jenbutongit:
reason for this PR:
DisciteCourseActions were handled by the template with a series of if statements. It became very hard to read and difficult to add any new statements in the correct places. DisciteCourseModules has less logic but good to keep it consistent anyway. Also makes it more testable. 🤓❤️ 

**PR includes:**
* [x] Refactor Course Actions into a .ts function
  * [x] Unit tests 
  * [x] should probably move this into lib/ui
* [x] Refactor Module Actions into .ts function
  * [x] Unit tests
* [x]  [LPFG-302](https://cshrdigitalandanalysis.atlassian.net/browse/LPFG-302) 'Just added' label
* [x] [LPFG-426](https://cshrdigitalandanalysis.atlassian.net/browse/LPFG-426) using correct attributes (download, target=_blank) for modules
* [x] squash!

needs fixing:
* [ ] integration tests

### DIFF
--- a/src/lib/learnerrecord.ts
+++ b/src/lib/learnerrecord.ts
@@ -123,11 +123,7 @@ export async function getReadyForFeedback(learningRecord: model.Course[]) {
 }
 
 export function isActive(record: CourseRecord) {
-	return (
-		record.state !== 'ARCHIVED' &&
-		record.state !== 'UNREGISTERED' &&
-		record.preference !== 'DISLIKED'
-	)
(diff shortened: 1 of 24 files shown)


m24murray:
Looks like tests are failing

jenbutongit:
I must've scrapped the class identifier the test was relying on when I was simplifying the .html file. I can add back or another way of verifying can be used. In the unit tests i'm checking if `/book/` is in the href string

put as a todo in OP

PR #364, opened 2018-06-27
LPFG-425: 'Ouch' page improvements (errors on non-production envs)

robleney-kainos:
* Show detailed error message or generic error message depending on environment. At the moment non-production environments are those with an ENV_PROFILE of 'dev' or 'test'. This may need to be changed.
* Added errorHandler with template for 500 errors and removed static 500.html error page.
* Removed handling for non 'text/html' requests as it was deemed unnecessary.
* Changed page directory in template.ts to use project root directory rather than cwd() to allow tests to pass (test and deployed application run from different directories).
* Replaced Jasmine with Mocha and Chai.


superafroman:
@williampri good if you could take a look at this

PR #368, opened 2018-06-29
Feature/lpfg-463

jenbutongit:
Sorry to leave this one hanging 😢. Feel free to ask Qs/discuss with me in comments after 29/06/2018.

[LPFG-463](https://cshrdigitalandanalysis.atlassian.net/browse/LPFG-463)
todo:
for all pages, check the copy is correct and styling.
- [x] Flow/ routing
- [x] Overview page
- [x] Date selection page
- [x] Accessibility page
  - Other description field
    - [x] put into template
    - [x] Spacing for 'other' box and text input
    - [x] Show other description field on the summary page
    - [x] Save other description field in session, if the user wants to change it at the summary stage it (navigate back to this part) it should persist
- [x] Pay by PO page
- [x] confirm PO page
  - [x] add to flow
- [x] Summary page
  - [x] after you make amends, it should take you back to the summary page rather than through the flow again. If it helps there is ?ref=summary query.

notes: 
* the flow should all be there now, just needs styling. Also with the PO flow I haven't checked what happens when you try the financial approver story, so that might be broken too. There is `?ref=summary` appended to the 'change' links on the summary page if that helps or you can get rid of it.
* Renamed confirmation to 'summary' because that's the Business Language™️. Didn't get round to changing around the classes though.. sorry!
* Book had a lot of routes so I've abstracted them out into `controllers/booking/routes.ts`. It's then called by server.ts like so: `app.use(bookingRouter.router)`
* styling docs, just in case


jenbutongit:
:( my beautiful PR has been soiled by changing base/incorrect merge. Under no circumstance merge this PR in it's current state otherwise you will add ~40 commits (which have already been added about 3 other times into **master**. You need to rebase SHA-1 6d67ca9 onto a new branch off dev (or master). 

cc: #376 #375

PR #382, opened 2018-07-13
feature/lpfg 571

superafroman:


williampri:
Optional flag always applied to newly created courses/modules, unable to set newly created courses to mandatory. Unable to set module as optional for existing learning (e.g. health and safety)

PR #387, opened 2018-07-20
chore/lpfg-586: update docker compose, readme, prettier formatting

m24murray:
- updating docker-compose to include new management app 🚀
- updating readme 📖
- **loads** of prettier formatting changes - we might want to make sure we're all using same config

note: will need updates to identity-service in separate PR for the client redirect_uri if trying to run lpg-management via docker compose


src/management-ui/page/courses/list.html
```diff
@@ -17,7 +17,7 @@ <h1 class="heading-large">Courses</h1>
                             <a href="/courses/{course.id}">{course.title}</a>
                         </th>
                         <td>
-                            <a href="{$config.LPG_UI_SERVER}/courses/{course.id}/">Review</a>
+                            <a href="{$config.LPG_MANAGEMENT_URL}/courses/{course.id}/">Review</a>
```
superafroman: This was originally put in as a 'preview' type thing - so view the course on LPG UI. No need to worry about though as new admin coming.

PR #396, opened 2018-08-08
feature/remove management-ui and update to new domain.

m24murray:
- removing old management directory 💣
- updating course and module models 🛠


src/lib/model.ts
```diff
@@ -15,7 +15,23 @@ export class Course {
...
+		course.modules = (data.modules || []).map(Module.create)
+
+		const audiences = (data.audiences || []).map(Audience.create)
+		course.audiences = audiences
+
+		if (user) {
+			let matchedAudience = null
+			let matchedRelevance = -1
+			for (const audience of audiences) {
+				const relevance = audience.getRelevance(user!)
```
jakub-kainos: exclamation mark redundant as there's `if(user)`

src/lib/model.ts
```diff
@@ -26,10 +42,12 @@ export class Course {
 	description: string
 	duration: number
 	learningOutcomes: string
-	price: number | string = this.collatePrice() || 'Free'
 
 	modules: Module[]
 
+	audiences: Audience[]
+	audience?: Audience
```
jakub-kainos: is that we don't have `audience?: Audience` anymore? (at least in `lpg-management` model)

src/lib/model.ts
```diff
@@ -123,28 +125,22 @@ export class Course {
...
 		const completionDate = this.getCompletionDate()
-		for (const module of this.modules) {
-			const moduleNext = module.nextRequiredBy(completionDate)
-			if (!next) {
-				next = moduleNext
-			} else if (moduleNext && moduleNext.getTime() < next.getTime()) {
-				next = moduleNext
-			}
+		if (this.audience) {
+			return this.audience!.nextRequiredBy(completionDate)
```
jakub-kainos: exclamation mark redundant and ternary operator would have been even leaner

src/lib/model.ts
```diff
@@ -172,27 +168,13 @@ export class Course {
 
 	shouldRepeat() {
 		const completionDate = this.getCompletionDate()
-		for (const module of this.modules) {
-			const moduleShouldRepeat = module.shouldRepeat(completionDate)
-			if (moduleShouldRepeat) {
-				return true
-			}
+		if (this.audience) {
+			return this.audience!.shouldRepeat(completionDate)
```
jakub-kainos: same comment about exclamation mark like above

src/lib/ui/moduleCallToAction.spec.ts
```diff
@@ -80,22 +80,23 @@ describe('Module call to actions', () => {
...
-						testUser,
-						'',
-						course.record!,
-						modifier
-					)
 					expect(cta.url).to.contain('#modules')
 				})
 
 				it('should return Add (to learning plan)', () => {
+					console.log(cta)
```
jakub-kainos: debug `console.log`?

PR #402, opened 2018-09-17
LPFG-768-update-learning-cat-to-oauth

m24murray:


docker-compose.yaml
```diff
@@ -116,17 +119,4 @@ services:
     ports:
       - 9004:9004
 
-#  lpg-management:
```
elxsy: Did you kill of management? accident?

elxsy:
we can optimize the changes but that's a task for another time. please see my previous comment

PR #418, opened 2018-10-15
Fix/LPFG-799-v2 Fixed bug

Peter-D-Taylor:
A fix that stops all required courses appearing as in every departments learning plan.

Fixed 'isInLearningPlan' being hard coded to true in courseControllerToAction.ts
Fixed audiences getting set to empty array in Course in model.ts

### DIFF
--- a/src/lib/model.ts
+++ b/src/lib/model.ts
@@ -17,11 +17,11 @@ export class Course {
 
 		course.modules = (data.modules || []).map(Module.create)
 
-	        let audiences = []
-        	if(data.course && data.course.audiences) {
-            		audiences = (data.course.audiences).map(Audience.create)
-        	}
-        	course.audiences = audiences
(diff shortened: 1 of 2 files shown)


src/lib/model.ts
```diff
@@ -17,7 +17,10 @@ export class Course {
 
 		course.modules = (data.modules || []).map(Module.create)
 
-		const audiences = (data.audiences || []).map(Audience.create)
+		let audiences = []
+		if (data.course && data.course.audiences) {
+			audiences = (data.course.audiences).map(Audience.create)
+		}
```
jakub-kainos: this looks wrong to me - audiences are at the same level as modules - so it should be they either both come from `data.course` or both from `data` - like it was originally

is there a test that could demonstrate this fix is required that way?

PR #431, opened 2018-11-05
LPFG-851: app insights

m24murray:
- adding app insights to lpg-services for azure (https://docs.microsoft.com/en-us/azure/application-insights/app-insights-nodejs)
- `do not merge` while Rob tests on Integration


m24murray:
I'm happy that if we do not have an instrumentation_key env var set, the app will still function as normal.

PR #432, opened 2018-11-06
Suggestions for you v2

mick-jones:
Changed home_page_suggestions value in en.json
Added new colour variable to the (previously empty???) varaibles.scss
Added styling for suggestions quick links
Mobile responsiveness added
Included an ID attribute to the H2 headings on SuggestedSection component
Updated course object to distinguish between primaryAreaOfWork and other areasOfWork
Added html for quick links in suggested.html

### DIFF
--- a/src/ui/assets/styles/components/_suggestions.scss
+++ b/src/ui/assets/styles/components/_suggestions.scss
@@ -1,4 +1,4 @@
-suggestions-wrapper {
+.suggestions-wrapper {
     width: 100%;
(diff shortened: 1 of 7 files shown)


src/ui/assets/styles/components/_suggestions.scss
```diff
@@ -0,0 +1,38 @@
+suggestions-wrapper {
```
m24murray: does this need a `.`

PR #436, opened 2018-11-08
LPFG-936: Manage organisations (updating docker-compose)

m24murray:
- updating docker-compose for csrs changes


m24murray:
not needed anymore

PR #444, opened 2018-11-22
LPFG-989: Update Learner UI to register learner

robleney-kainos:
Requires learner-record running branch LPFG-989 to run locally.


m24murray:
Just need to rebase @robleney-kainos

PR #445, opened 2018-11-23
LPFG-975: Add accessibility details to booking

robleney-kainos:
Updated lpg-ui to use notification service insted of gov notify.


src/lib/feedback.ts
```diff
@@ -1,24 +1,19 @@
 import * as config from 'lib/config'
-import * as gov from 'notifications-node-client'
+import {NotificationService} from '../notification-service'
+import {NotificationServiceConfig} from '../notification-service/notificationServiceConfig'
+// import * as gov from 'notifications-node-client'
```
m24murray: Remove comment

src/lib/feedback.ts
```diff
@@ -1,24 +1,19 @@
...
-	const notify = new gov.NotifyClient(config.GOV_NOTIFY_API_KEY)
+},                           accessToken: string) {
+	const notify = new NotificationService(new NotificationServiceConfig())
 	for (const recipient of config.FEEDBACK_RECIPIENTS) {
-		const resp = await notify.sendEmail(
+		await notify.sendEmail(
 			config.FEEDBACK_TEMPLATE_ID,
 			recipient,
-			{personalisation: req}
+			req, accessToken
```
m24murray: This formatting looks a little weird to me, could you check?

src/ui/controllers/feedback.ts
```diff
@@ -14,7 +14,7 @@ export async function record(req: express.Request, res: express.Response) {
 		const param = {pageUrl, wentWrong, whatDoing}
 		try {
 			// const a = 2
-			await feedback.record(param)
+			await feedback.record(param, req.user.accessToken)
```
m24murray: While we're here, can we get rid of that comment above?

robleney-kainos:
notification-service was under /src similar to lpg-management. I guess I'm clinging on to the hope that maybe it will be a separate library. I've moved it to /src/lib/service.

m24murray:
Replaced by https://github.com/Civil-Service-Human-Resources/lpg-services/pull/536

PR #449, opened 2018-11-27
[LPFG-1039][BUG] New orgs showing as undefined

elxsy:


elxsy:
breaking so much stuff

PR #451, opened 2018-11-28
Lpfg 972 attendant tracking

Peter-D-Taylor:
Added availability to Venue in model.ts
Added availability column and hide radio button when course is fully booked to choose-date.html


src/ui/page/booking/choose-date.html
```diff
@@ -41,22 +42,28 @@ <h2 class="heading-medium u-space-t10 heading--red-top">{ module.title || course
...
+                                            <label for="radio-{event.id}" class="multiple-choice__label">
+                                                <span class="multiple-choice__spanner">
+                                                    <span class="visuallyhidden">{$datetime.formatDate(event.date)}, {event.location}</span>
+                                                </span>
+                                            </label>
+                                        </div>
                                     </div>
-                                </div>
+                                {:else}
+                                    <div class="div-table__cell">fully booked</div>
```
m24murray: Can we capitalise this to sentence case?

Peter-D-Taylor: The design has it as lower case, but I could change it if you think that would be better

PR #459, opened 2018-11-30
LPFG-1053 Learning record text

mick-jones:
- Added learning record help text
- Added additional styling to header


src/ui/page/learning-record/index.html
```diff
@@ -1,6 +1,7 @@
 <Page title="Learning record" {_csrf}>
     <div class="container">
-        <h1 class="heading-xlarge">Learning record</h1>
+        <h1 class="heading-xlarge small-margin">Learning record</h1>
+        <p class="govuk-body">This is a record of the learning you have completed.If a course has modules it will only appear in your learning record when all its modules have been completed.</p>
```
m24murray: Space between full stop and 'If'.

Also I do not believe the semantics of that message are correct, have a question out with James to confirm.

PR #461, opened 2018-11-30
LPFG-1056 Change casing of grades

mick-jones:
- Changed the case of grades entries in en.json


m24murray:
This will be done via db

PR #463, opened 2018-12-03
LPFG-1059 new header & text for add line manager email page

mick-jones:
- Changed header text for add/edit line manager page


mick-jones:
This branch seems very broken - deleting

PR #464, opened 2018-12-03
LPFG-1058 Suugested learning removal message change

mick-jones:
- changed flash message when user removed suggested learning


wisdome123:
this change is in the branch LPFG-1059
https://github.com/Civil-Service-Human-Resources/lpg-services/pull/463

PR #480, opened 2018-12-20
LPFG-1058 Suggested course removal text

mick-jones:
- Changed text when removing a course from suggested learning


m24murray:
@mick-jones can you look at conflicts?

PR #481, opened 2019-01-04
Lpfg 669 event status

Peter-D-Taylor:
**Needs to be merge with LPFG-669 from learner-record and learning-catalogue**

This hides events that are cancelled so that learners can't book onto them.

- model.ts: Added status to Event, if the status is null then the event is set to active so that old events will till appear.
- choose-date.html: hide events that are cancelled

### DIFF
--- a/src/lib/model.ts
+++ b/src/lib/model.ts
@@ -287,10 +287,7 @@ export class Event {
 			capacity = data.capacity
 		}
 
-		let status = 'Active'
-		if (data.status) {
-			status = data.status
-		}
+		const status = data.status ? data.status : 'Active'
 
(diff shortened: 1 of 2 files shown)


src/lib/model.ts
```diff
@@ -287,23 +287,30 @@ export class Event {
 			capacity = data.capacity
 		}
 
-		return new Event (date, location, capacity, availability, data.id)
+		let status = 'Active'
+		if (data.status) {
+			status = data.status
+		}
```
m24murray: Could refine this a bit with a ternary:

```suggestion
		 status = data.status ? data.status : 'Active'
```

PR #493, opened 2019-01-17
LPFG 1172 cancellation reason message

Peter-D-Taylor:
Get cancellation reasons from learner record, send one back with cancellation, to send in the email and save in the db.


src/lib/learnerrecord.ts
```diff
@@ -72,6 +73,18 @@ export async function bookEvent(
...
+	try {
+		return await http.get(`/event/booking/userCancellationReasons`, {
+			headers: {
+				Authorization: `Bearer ${user.accessToken}`,
+			},
+		})
+	} catch (e) {
+		logger.error(e)
+	}
+}
```
m24murray: Can you handle this await using then/catch, instead of a try/catch? If an error occurs getting the cancellation reasons, the app will just hang.

PR #509, opened 2019-01-25
LPGF-1230 property labels

mick-jones:
- Added property labels for all course and module property fields


src/ui/component/disciteParts/DisciteModuleProperties.html
```diff
@@ -2,14 +2,17 @@
...
             <li class="discite__property-item">
-                <span class="visuallyhidden">Duration:</span> {module.duration || ''}
+                <span>Duration:</span>
+                <span class="lpg-course-duration"> {module.duration || ''}</span>
             </li>
             {#if module.type !== 'file'}
                 <li class="discite__property-item">
-                    <span class="visuallyhidden">Cost:</span> &pound;{module.cost || 'Free'}
+                    <span>Cost:</span>
+                    <span class="lpg-course-cost">&&pound;{module.cost || 'Free'}</span>
```
Peter-D-Taylor: Looks like there's an extra '&' here, also should there be a pound sign if the module is free?

mick-jones: Good catch on the ampersand, but no, the pound signs have been removed from free courses

mick-jones: removed

PR #515, opened 2019-01-28
LPFG-1246 Required learning

mick-jones:
- added extra text in the event that a learner has either, no courses left in their required learning, or, still has courses to complete. If the latter, there is a link to the required learning, homepage


laurentaylor1:
Tested on branch:

Incorrect value for "and have **3** more to do." The total remaining courses is actually 2, but it is displaying the total number of courses, regardless of completion.
Screen Shot 2019-03-29 at 16 14 38

laurentaylor1:
Successfully test on branch:

2 total courses:
Screen Shot 2019-03-29 at 15 58 44

1 new course added - 1 completed, 2 remaining to complete:
Screen Shot 2019-03-29 at 16 31 06

No courses completed, 1 to be completed:
(TBC)

laurentaylor1:
**Resolved**: Pushed logic change for "You have completed all your required learning for this reporting year.":
Screen Shot 2019-04-01 at 10 32 26

All testing completed.

PR #518, opened 2019-01-29
LPFG-1235 Mobile search

mick-jones:
- added show/hide functionality for mobile search filters.


laurentaylor1:
Tested on branch: Padding should be added to the bottom of the check boxes, as they are different from the Sketch prototype, see images below.

Branch:
Screen Shot 2019-03-29 at 13 43 49

Prototype:
Screen Shot 2019-03-29 at 13 44 56

PR #523, opened 2019-01-31
LPFG-1269 Discite properties width

mick-jones:
- fixed the width on the discite properties module that was causing the text to drop below the bottom border


laurentaylor1:
Screen Shot 2019-03-26 at 15 57 43
Successfully tested on integration.

PR #526, opened 2019-02-05
Update dc and admin nav

m24murray:


src/ui/component/Nav.html
```diff
@@ -21,17 +21,12 @@
             <li class="active"><a href="/profile">Profile</a></li>
         {:else}
             <li><a href="/profile">Profile</a></li>
-        {/if} 
-        {#if $req.originalUrl.includes("/skills")}
```
Peter-D-Taylor: Was this supposed to be deleted?

m24murray: Yes will be coming with rest of skills work

PR #555, opened 2019-04-08
LC-182: Search filters refactor

laurentaylor1:
- title and arrowhead should be vertically centred in each grey box
- "+" should be in-line with filter title


laurentaylor1:
Replacement PR has been created for this code:
https://github.com/Civil-Service-Human-Resources/lpg-services/pull/560

PR #596, opened 2019-10-29
Feature session  in redis

michalpiatek:


src/lib/ui/courseCallToAction.spec.ts
```diff
@@ -34,13 +34,13 @@ const moduleData: {[module: string]: any} = {
 		events: [
 			{
 				capacity: 30,
-				dateRanges: [{date: '2019-10-10', startTime: '06:00:00'}],
+				dateRanges: [{date: '2020-10-10', startTime: '06:00:00'}],
```
charlesluokb: `2019-10-10` would be right as id says `past`?

PR #598, opened 2019-11-06
Added a configurable maintenance page (similar to the one on identity

michalpiatek:
service)


src/ui/page/maintenance.html
```diff
@@ -0,0 +1,1179 @@
...
+    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-22141655-4">
+    </script>
+    <script>
+        window.dataLayer = window.dataLayer || [];
+        function gtag() { dataLayer.push(arguments); }
+        gtag('js', new Date());
+        gtag('config', 'UA-22141655-4', { anonymize_ip: true });
+    </script>
+    <style>
```
charlesluokb: can we move the style configurations (i.e. scss stuff) to a separate file to make this page more readable?

src/ui/server.ts
```diff
@@ -20,12 +20,14 @@ import * as i18n from 'lib/service/translation'
 import {ProfileChecker} from 'lib/ui/profileChecker'
 import * as template from 'lib/ui/template'
 
+import * as maintenanceMiddleware from 'lib/service/maintenance'
```
charlesluokb: as per naming convention, I think people are more familiar with `xxxxService` :-) can we call it `maintenanceService` ? be in line with its parent folder name i.e. `lib/service`.

charlesluokb:
I have not seen spec tests (under /webdriver/src/test) for the maintenance page. are they in different location? can we have them written? @michalpiatek

src/ui/server.ts
michalpiatek: sure

src/ui/page/maintenance.html
michalpiatek: it is a copy & paste from error.html page - I can amend

m24murray:
Closing this as no longer needed.

PR #599, opened 2019-11-10
Profile page improvements

michalpiatek:
* Added userId so that usage of HATEOAS is reduced (fewer calls)
* unified axios and traverson handler in patchSivilServant handler
* Bumped version of appInsights to 1.5.0 and change order of initialization so that appInsights will track the dependencies now

### DIFF
--- a/src/ui/controllers/user.ts
+++ b/src/ui/controllers/user.ts
@@ -198,6 +198,147 @@ export function parseRoles(traversonResult: any): [Level[], string] | void {
 	}
 }
 
+export async function newRenderAreasOfWorkPage(
+	req: express.Request,
+	res: express.Response
+) {
+	const lede = req.__('register_area_page_intro')
+	let selectedArr: any = []
+	let currentLevel: number = 0
+	let selected: number
+	let levels: Level[][] = []
+	let prevLevelUrl
+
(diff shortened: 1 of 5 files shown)


src/ui/controllers/user.ts
```diff
@@ -529,7 +391,7 @@ export async function patchAndUpdate(
 				res.redirect(`/profile/${inputName}`)
 			})
 		}
-	} else if (response) {
+	} else if (requestWasSuccessful) {
 		// seems like we have to get the profile again to get values
```
charlesluokb: just a question, are we going to do something about this unnecessary call to get `profile`?

src/ui/controllers/user.ts
```diff
@@ -193,147 +194,6 @@ export function parseRoles(traversonResult: any): [Level[], string] | void {
 	}
 }
 
-export async function newRenderAreasOfWorkPage(
```
charlesluokb: can you explain why this function and one below are removed? are they not referenced?

src/ui/controllers/user.ts
michalpiatek: it is currently necessary as the patch method returns the civilServant in the HATEOAS format

PR #600, opened 2019-11-13
Moved appInsights initialization to the top of server startup script

michalpiatek:


src/lib/registry.ts
```diff
@@ -20,9 +20,8 @@ traverson.registerMediaType(hal.mediaType, hal)
 export async function get(node: string): Promise<{}> {
 	const result = await new Promise((resolve, reject) =>
 		traverson
-			.from(config.REGISTRY_SERVICE_URL)
+			.from(config.REGISTRY_SERVICE_URL + '/' + node)
```
m24murray: If this is for performance, is it maybe worth replacing traverson/hal with a more basic HTTP request using something like Axios? I'm pretty sure that I've replaced traverson requests in a few places using the getWithoutHal function at the bottom of this class.

michalpiatek: removed it from the PR - not part of the change

PR #609, opened 2020-01-31
perf: updating maxSockets and maxFreeSockets config for HTTP

m24murray:
- updating maxSockets and maxFreeSockets config for HTTP as per MS request


m24murray:
@willh - changed base branch as requested

PR #630, opened 2020-04-09
[lc-431] - Remove HATEOAS pattern from search functionality, replace …

mattdoherty-kainos:
…with direct calls to requried CSRS endpoints, performance fix


willh:
Good to go into release F after positive test results

PR #631, opened 2020-04-10
chore: fixing typescript warnings

m24murray:
TS/node versions updated on the app. ts-igoring some of the warnings as we know these already work, and do not want to explicitly cast

### DIFF
--- a/src/ui/controllers/booking/booking.ts
+++ b/src/ui/controllers/booking/booking.ts
@@ -97,6 +97,7 @@ export async function renderChooseDate(
 		req.flash('bookingSelected')[0] || req.session!.selectedEventId || null
 
 	if (!selectedEventId) {
+		// @ts-ignore
 		selectedEventId = req.query.eventId
 		if (req.session!.selectedEventId) {
 			req.session!.selectedEventId = selectedEventId
--- a/src/ui/controllers/home.ts
+++ b/src/ui/controllers/home.ts
@@ -100,6 +100,7 @@ export async function home(req: express.Request, res: express.Response, next: ex
 		let noOption
 
 		if (req.query.delete) {
+			// @ts-ignore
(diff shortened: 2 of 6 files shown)


m24murray:
Hi @willh,

Smoke and regression tested these changes on integration, specifically focusing on areas impacted by the 6 file changes included in this PR.

These are as follows:
* src/ui/controllers/booking/booking.ts
	* Render choose date screen
* src/ui/controllers/home.ts
	* Moved a booked course out of learning plan into Learner record
	* Removed a course from learning plan
* src/ui/controllers/learning-record/index.ts
	* Take and complete new learning
* src/ui/controllers/search.ts
	* Search
* src/ui/controllers/user.ts
	* Update area of work on profile
* src/ui/controllers/xapi.ts
	* Take and complete elearning package

image
image
image
image
image
image
image

Standard smoke test and regression test of these areas completed successfully with no concern.

willh:
This one is a tricky one. Comparing the two identical build runs of the head of master:
http://lpgjenkins.uksouth.cloudapp.azure.com/blue/organizations/jenkins/lpg-services-pipeline/detail/master/25/pipeline/ and http://lpgjenkins.uksouth.cloudapp.azure.com/blue/organizations/jenkins/lpg-services-pipeline/detail/master/24/pipeline/ - in particular the version control checkout stage showing the checking out of commit ID `f8ce4a0b2cf55fdd0b4555da6246d8c2431054e1`. It is clear that two  builds of the same codebase have two different results. 

The difference is in packages resolved at build time that have updated dependencies which cause the false negative in type errors, so until a general exercise of updating dependencies and regression testing takes place (and should take place as a regular chore every few months) it is best to `ts-ignore` annotate the source after the additional manual regression test has proven along with the unit tests that the functionality still behaves as expected.

We should also recommend to cherry-pick this commit into other active branches until the F2F release goes live and merges into master.

PR #641, opened 2020-04-15
LC-291 fixing course which has a F2F module with no event

KLV96:
Page breaks so we added an if statement to check if the events length is bigger than 0 and if the events is not null

### DIFF
--- a/src/lib/model.ts
+++ b/src/lib/model.ts
@@ -118,27 +118,29 @@ export class Course {
 				if (module.events[0]) {
 					const event = module.events[0]
 					let durationInSeconds = 0
+					if (event.dateRanges) {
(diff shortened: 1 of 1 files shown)


src/lib/model.ts
```diff
@@ -118,27 +118,29 @@ export class Course {
 				if (module.events[0]) {
```
kameshganesan, line 118: what if there are more than 1 event?

KLV96: I think it will not be a problem as this checks if there is first event then it means that events array is not empty and we can look through it. Shall I change it to the check null and the check length?

KLV96: after having a chat with Kamesh, I've changed the if statement and pushed again

PR #649, opened 2020-04-29
LC-14: refactoring change your email

m24murray:
Refactoring change your email flow. If a user doesn't have an organisation, we should just send the user to the change email form because we cant check whether they are an agency token user. 

Furthermore, this was only handling users that were either Whitelisted or Agency tokens, but this did not handle users who had been invited via IDM. Updating logic accordingly.

Adding model and factory classes for agency token objects.

Lots of prettier formatting updates.

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -4,5 +4,6 @@
 	"semi": false,
 	"singleQuote": true,
 	"trailingComma": "es5",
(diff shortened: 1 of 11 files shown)


mattdoherty-kainos:
Reviewed, looks good.

PR #650, opened 2020-05-05
Changing Day to be 8 hours instead of 7

KLV96:
I've changed a day to be 8 hours as requested by the client and deleted a function which was not being used. I also changed the test cases so that it calculated a day being 8 hours


.gitignore
```diff
@@ -1,14 +1,12 @@
-**/*.iml
 *.log
-**/*.iml
+*.iml
```
kameshganesan, line 2: revert this change

KLV96: *.iml ignores the top level .iml file (which is only one we have in our structure). This change is from the commit that Matt Doherty had. So,  I will keep it to keep the .gitignore file consistent between branches.

PR #651, opened 2020-05-06
Lc 391 required suggested learning issues

PortobelloRain:
Few front end changes for Required/Suggested Learning

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -70,7 +70,7 @@ services:
     image: lpgregistry.azurecr.io/identity-service:master-4
     ports:
       - 8080:8080
-      - 8081:8081
+#      - 8081:8081
 
   learner-record:
     container_name: learner-record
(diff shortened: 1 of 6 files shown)


willh:
Can you try forcing Jenkins to rebuild (that should clear cached files as a pre-requisite step to npm install) and see if it still happens?

PortobelloRain:
That cleared the cache but still getting the following errors on jenkins npm run build:

src/lib/config/passport.ts:109:17 - error TS4058: Return type of exported function has or is using name 'QueryString.ParsedQs' from external module "/mnt/workspace/ired-suggested-learning-issues_2/node_modules/@types/qs/index" but cannot be named.



109 export function hasRole(role: string) {

                    ~~~~~~~





src/lib/config/passport.ts:122:17 - error TS4058: Return type of exported function has or is using name 'QueryString.ParsedQs' from external module "/mnt/workspace/ired-suggested-learning-issues_2/node_modules/@types/qs/index" but cannot be named.



122 export function hasAnyRole(roles: string[]) {

                    ~~~~~~~~~~





src/lib/ui/profileChecker.ts:34:2 - error TS4053: Return type of public method from exported class has or is using name 'QueryString.ParsedQs' from external module "/mnt/workspace/ired-suggested-learning-issues_2/node_modules/@types/qs/index" but cannot be named.



34  checkProfile() {

willh:
That seems similar to the issue that pushed us towards adding some of those @ts-ignore annotations - if you suppress the type warning does it function as expected anyway? When we added the other ignore annotations we had to regression test to be sure it was just a type mismatch caused by the updated library version and not an actual semantic change in the data structures. I think you have a couple of options around either ts-ignore annotating them or updating the import from that library reference, but I'm not close enough to the detail.

PR #652, opened 2020-05-06
[lc-536] - Update CSP to have correct data source, removed single quo…

mattdoherty-kainos:
…tes around data: entry


willh:
Target branch TBC depending on release - possibly one for F2F

PR #659, opened 2020-06-01
LC-337: removing forceOrgChangeFlag, updating change email flow

brdgeth:
Updates to change email flow have reduced the complexity of the changeEmail function on lpg-services and now means it is agnostic of whether the user is whitelisted, invited or an agency user.

The forceOrgChange flag is now redundant as we are now clearing the users organisation on change email as opposed to setting a flag value. Attribute removed from domain objects and service classes.

Updating respective tests.


m24murray:
Replaced by https://github.com/Civil-Service-Human-Resources/lpg-services/pull/662

PR #665, opened 2020-06-12
Bug fix lc 14 empty org error v3

brdgeth:
Fix for response already committed error which causes the UI to fail.


mattdoherty-kainos:
Merging

PR #668, opened 2020-06-24
LC-489-Learner Journey Completed with full coverage and extensive unit tests

amobarak:
Full learner journey complete, with unit tests and coverage, removed mocked responses and connected to backend, works with Civil-Servant-Registry-Service branch name `feature-LC469-Add-a-new-question`.


src/ui/controllers/skills/index.ts
```diff
@@ -1,143 +1,243 @@
...
-	// @ts-ignore
-	if (user.areasOfWork && user.areasOfWork.length > 0) {
-		show = user.areasOfWork[1].toLocaleLowerCase() === 'policy'
+	if (!(user.areasOfWork && user.areasOfWork.length > 0) || !user.organisationalUnit!.id) {
+		showQuizScreen(req, res, false)
+		return
 	}
-	res.send(template.render('skills/introduction', req, res, {show}))
+
+	const professionId = parseInt(user.areasOfWork[0], 10)
```
kameshganesan, line 25: do you have access to civil servant object, if not you can fire a get and grab one. That will have profession and organisation information. a get to this csrs endpoint `/civilServants/me`  will get the needed. then you can carry it through the journey.

src/ui/page/skills/questions.html
```diff
@@ -15,32 +30,23 @@ <h2 class="heading heading-large ">Question {index+1} of {count}</h2>
...
+                </form>
+                <form id="mainForm" action="skills/questions/{index}" method="post" class="question-form">
                     <input type="hidden" name="_csrf" value="{_csrf}">
                     <fieldset class="form-group">
-                        <QuizQuestion question={question} type={type} />
+                        <QuizQuestion question={question} keys={keys} skipped="{skipped}" attemptedAnswers="{answersToQuestionKeys}" multipleAnswers="{multipleAnswers}"/>
                     </fieldset>
                     <div class="dont-know">
-                        <a href="/skills/questions/{index+1}">I'm not sure</a>
+                        <a class="skipQuestion" onclick="$('#skip').submit()">Skip this question</a>
```
kameshganesan, line 42: What happens when JS is turned off?

src/ui/controllers/skills/index.ts
amobarak: The user object is exactly that, it calls "/civilServants/me" and saves it to the user object, and it contains all the information (profession/organisation), it's commonly used in different places in the app too.

src/ui/page/skills/questions.html
amobarak: They can click on Continue button, and that will function the same

amobarak: Offcourse, having JS turned off is not ideal and will cause other pages to act weird

src/ui/controllers/skills/index.ts
kameshganesan: Yes. I agree. It's a pity we need to access the professionId from an array. It would have been nicer if we were able to, say, access it like user.profession.primary. We can resolve this discussion.

src/ui/page/skills/questions.html
kameshganesan: hmm. hoping that doesn't become an issue later on. resolving discussion now.

PR #697, opened 2020-08-27
LC-190: accessibility update for search page

m24murray:
On mobile view, the search page will show a hide/show search filters checkbox, not visible during standard desktop view. Adding an aria-hidden = "true" to the element so that it is not narrated by screen readers.


mattdoherty-kainos (approved):
lgtm

PR #703, opened 2020-09-21
hotfix: remove tab index for main content

m24murray:


willh (approved):
Aligns with latest guidance on GDS design system

PR #707, opened 2020-10-30
LC-728-search-filters-accessibility

PortobelloRain:
Adding accessibility to search filters in LPG and fixing video mp4 in display-video.htm


neeraj-jain-cbo:
Reviewed the code with Vince.

PR #769, opened 2021-08-25
LC-1054 course status on home page and course details page for new learning period

neeraj-jain-cbo:


src/lib/learnerrecord.ts
```diff
@@ -319,6 +320,45 @@ export class CourseRecord {
...
+				) {
+					completionDate = moduleRecord.completionDate
+				}
+			}
+			return completionDate
+		}
+		return undefined
+	}
+
+	//LC-1054: Below method is no longer used
```
jaylco, line 361: This should work the same but has more streamlined logic:
`	getCompletedModules() {
		// Return only the modules that are completed (where m.completionDate is not null)
		return this.modules.filter(m => m.completionDate)
	}

	//LC-1054: Rather than renaming the above method a new method is Implemented as below
	getLatestCompletionDateOfModulesForACourse() {
		if (this.isComplete()) {
			let completedModules = this.getCompletedModules()
			if (completedModules.length > 0) {
				let completedDates = completedModules.map(m => new Date(m.completionDate || 0))
				return _.max(completedDates)
			}
		}
		return undefined
	}

	//LC-1054: A new method implemented as below
	getEarliestCompletionDateOfModulesForACourse() {
		if (this.isComplete()) {
			let completedModules = this.getCompletedModules()
			if (completedModules.length > 0) {
				let completedDates = completedModules.map(m => new Date(m.completionDate || 0))
				return _.min(completedDates)
			}
		}
		return undefined
	}`

jaylco (commented):
Left 1 comment

jaylco (changes requested):
https://github.com/Civil-Service-Human-Resources/lpg-services/pull/769/files#r700191295

src/lib/learnerrecord.ts
neeraj-jain-cbo: Review comments applied as below:

getCompletedModules() {
		// Return only the modules that are completed (where m.completionDate is not null)
		return this.modules.filter(m => m.completionDate)
	}

	//LC-1054: Rather than renaming the above method a new method is Implemented as below
	getLatestCompletionDateOfModulesForACourse() {
		if (this.isComplete()) {
			return _.max(this.getCompletedModules().map(m => m.completionDate))
		}
		return undefined
	}

	//LC-1054: A new method implemented as below
	getEarliestCompletionDateOfModulesForACourse() {
		if (this.isComplete()) {
			return _.min(this.getCompletedModules().map(m => m.completionDate))
		}
		return undefined
	}

jaylco (approved):

PR #826, opened 2022-10-24
Add optional UID column to ModuleRecord

SavvasCO:


SavvasCO:
## Integrationt test

Progress status gets updated in the frontend:

Screenshot 2022-10-24 at 13 35 56

PR #830, opened 2022-11-10
LC 1090 optimise org cache

jaylco:
- New classes/methods for handling the organisational unit cache. This cache will allow the frontend to fetch organisations without having to go to the CSRS application. Fetching the hierarchy is the most prominent use case for this, other use cases include
    - Fetching the organisational dropwdown (AKA typeahead) to allow users to select their department (either on signup or profile change)
    - Fetching the organisations for search filtering


src/lib/service/civilServantRegistry/csrsService.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+	const org = new OrganisationalUnit()
+	org.name = orgName
+	org.id = id
+	if (parentId) {
+		org.parentId = parentId
+	}
+	return org
+}
+
+describe('CsrsService tests', () => {
```
SavvasCO, line 20: Might be a good idea to write a few tests for `getOrganisation` covering the cases in your `if` statements (eg. `org === undefined`). Probably easier if you mock the response from `getOrganisationFromApi` and `organisationalUnitCache.get`.

SavvasCO (commented):
A c

src/lib/service/civilServantRegistry/csrsService.spec.ts
jaylco: Resolved - copied the tests over from the lpg-management PR (the code is identical more or less)

SavvasCO:
Looks good!

SavvasCO (approved):

PR #858, opened 2023-02-09
LC 1499 launch E-Learning via CSL-Service

jaylco:
- E-Learning launch is now handled by the CSL-Service API. The API returns a launch link which the frontend then passes onto the browser via 302


neeraj-jain-cbo (approved):
LGTM

neeraj-jain-cbo (dismissed):
LGTM

neeraj-jain-cbo (approved):
LGTM

PR #860, opened 2023-02-23
LC 1716 manage learne record API calls

jaylco:
- Prevent statements from going to learning locker
- Only update E-Learning on launch and COMPLETED verb
- Check if user is undefined and return if so, preventing the exception that occurs
- Wrap the learner record sync in a try catch to prevent container crashes


neeraj-jain-cbo (approved):
LGTM

PR #890, opened 2023-04-20
Fix course completion bug

SavvasCO:
This change:

* Filters out optional modules and deleted modules when calculating completion dates for a course

### DIFF
--- a/src/lib/learnerrecord.ts
+++ b/src/lib/learnerrecord.ts
@@ -327,8 +327,8 @@ export class CourseRecord {
 	}
 
 	//LC-1054: Rather than renaming the above method a new method is Implemented as below
-	async getLatestCompletionDateOfMandatoryModulesForACourse(user: model.User) {
-		const availableCompletedMandatoryCourses = await this.getAvailableCompletedMandatoryModulesForCourse(user)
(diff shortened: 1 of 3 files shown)


src/lib/learnerrecord.ts
```diff
@@ -357,6 +361,21 @@ export class CourseRecord {
...
+	async getAvailableCompletedMandatoryModulesForCourse(user: model.User) {
+		const availableModuleIds = await this.getAvailableModuleIdsForCourse(user)
+
+		return this.getCompletedModules()
+			.filter(module => module.optional === false)
+			.filter(module => availableModuleIds.includes(module.moduleId))
+	}
+
+	async getAvailableModuleIdsForCourse(user: model.User) {
+		const courseFromCatalog = await catalog.get(this.courseId, user)
```
jaylco, line 374: Could we use on the course data from the top-level `home.ts home` function for this? Seems like this will introduce a `GET /course/{courseID}` catalogue API call for each course processed?

SavvasCO: Good point! 👀

jaylco (approved):

PR #900, opened 2023-05-26
LC-2063 learning record bug

jaylco:
- `learning-record` now uses the required learning audience to calculate how far along a user is with their mandatory learning
- Implemented `getDisplayStateForCourse` which can be used for the homepage at a later date


neeraj-jain-cbo (approved):
LGTM

PR #906, opened 2023-06-15
LC 2102 homepage extra module fix

jaylco:
- Fixed homepage logic. Homepage now respects all relevant modules in a course and calculates completion statuses according to learning period. Refactored the `courseRecord` class to achieve this.
- Introduced a `CourseRcd` interface to ensure compatibility with legacy courseRecord and current courseRecord with a long-term aim to move away from the legacy `courseRecord` and `getRawLearnerRecord` code
- Added new `npm` command for running with just the TS compiler

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
 		"sass": "gulp sass",
 		"start:all": "npm run start:ui",
 		"start:build": "npm run build && npm run start:ui",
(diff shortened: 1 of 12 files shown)


neeraj-jain-cbo (approved):
LGTM

PR #916, opened 2023-07-27
LC-2258 suggested learning grade code bug

jaylco:
- `getGradeCode` - a new method on the user class which returns an empty string if the user's grade is null
- Suggestions for you now make use of the new above method to calculate suggested learning, even when the grade is null


neeraj-jain-cbo (approved):
LGTM

PR #997, opened 2024-11-25
LC-3210: Accessibility: Add description lists to profile page

SavvasCO:
This change replaces tables in the Profile page to description links to improve accessibility


views/assets/styles/main.css.map
jaylco, line 1: May need to go back and delete this so it registers in the .gitignore

jaylco (commented):
LGTM, 1 comment

views/assets/styles/main.css.map
SavvasCO: Fair point

PR #1012, opened 2025-01-14
LC-3374: User can remove all interests from profile

SavvasCO:
This change:

* Removes the restriction to allow interests from user profile to be empty
* Updates the profile page UI so that when no interests are selected, a "No interests selected" text appears
* Updates the Suggestions for You page so that when no interests are selected, a "No interests selected" text appears
* Updates the CSRS `PATCH` logic so that it can send empty interests list to the endpoint.
* Updates the unit tests to check if the user now gets redirected when no interests are selected

### DIFF
--- a/src/ui/controllers/profile/models/interests/editInterestsPageModel.ts
+++ b/src/ui/controllers/profile/models/interests/editInterestsPageModel.ts
@@ -1,9 +1,5 @@
-import {ArrayNotEmpty} from 'class-validator'
 import {CreateInterestsPageModel} from './createInterestsPageModel'
 
 export class EditInterestsPageModel extends CreateInterestsPageModel {
-	@ArrayNotEmpty({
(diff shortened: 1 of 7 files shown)


src/ui/controllers/profile/models/interests/editInterestsPageModel.ts
```diff
@@ -1,9 +1,5 @@
-import {ArrayNotEmpty} from 'class-validator'
 import {CreateInterestsPageModel} from './createInterestsPageModel'
 
 export class EditInterestsPageModel extends CreateInterestsPageModel {
-	@ArrayNotEmpty({
-		message: 'profile.interests',
```
jaylco, line 6: Could we also remove the entry in the `en.json` for this please

jaylco (approved):

PR #1020, opened 2025-02-18
LC-3213 Search accordion accessibility

jaylco:
- Use buttons for the search page accordions
- Add aria attributes to accordions
- Remove JQuery code for accordions


views/component/LeftSearchBox.html
```diff
@@ -8,110 +8,140 @@
         <div class="lpg-related-items form-group govuk-related-items--search leftSearchBox filter-toggle">
             <fieldset>
                 <legend>
-                    <h2 class="govuk-related-items__heading heading-small filter-headings" tabindex="0">Learning type</h2>
+                    <h2 class="govuk-related-items__heading">
+                        <button class="filter-headings" type="button" aria-controls="accordion-learning-type" aria-expanded="true" aria-label="Learning type" tabindex="0">Learning type</button>
```
SavvasCO, line 12: Could we possibly also add a `aria-haspopup="true"` to the buttons? It's an optional attribute but it will announce to the screen reader users that [this button has a menu](https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/expanded/#adding-haspopup-optional).

SavvasCO (commented):
Looks great. A minor recommendation.

views/component/LeftSearchBox.html
jaylco: Hmmm, that article suggests "We do not recommend to use `area-haspopup` [alongside `aria-expanded`] as it does not provide any additional value"

SavvasCO: Fair enough.

SavvasCO (approved):

PR #1021, opened 2025-02-19
LC-3224 button focus non-text contrast

jaylco:
- Fix non-text contrast by bringing green CTA button's focus state inline with GDS buttons: https://design-system.service.gov.uk/components/button/
- Fix other accessibility issue where names and ids for profile checkboxes/radios had spaces in


SavvasCO (approved):
LGTM

PR #1024, opened 2025-03-06
LC-3431: Show correct event status

SavvasCO:


src/lib/service/learnerRecordAPI/event/service.ts
```diff
@@ -0,0 +1,44 @@
...
+export async function getCancelledEventUidsFromCourseRecord(courseRecords: CourseRecord | CourseRecord[], user: User) {
+	let records: CourseRecord[]
+	if (Array.isArray(courseRecords)) {
+		records = courseRecords
+	} else {
+		records = [courseRecords]
+	}
+
+	const eventIds = getEventUidsFromCourseRecords(records)
+	const learnerRecordEvents: Event[] = await getEventsByUids(eventIds, user)
```
jaylco, line 17: We shouldn't be accessing the API if `eventIds` is empty. This is causing an exception when I load the homepage with nothing in my learning plan

jaylco (changes requested):

neeraj-jain-cbo (approved):

jaylco (approved):

PR #1103, opened 2026-07-03
LC-4278: New header

SavvasCO:


views/component/Header.html
```diff
@@ -3,29 +3,33 @@
...
-                    </svg>
-                    GOV.UK
+                <a href="/login" title="Go to the GOV.UK homepage" id="logo" class="content">
```
jaylco: `/login` is a dead link on lpg-services - could we just use the `https://www.gov.uk` for now please

views/component/Header.html
```diff
@@ -3,29 +3,33 @@
...
+                            </svg>
+                            <div class="gov-uk-name">GOV.UK</div>
+                        </div>
+                        <div class="service-name">Civil Service Learning</div>
```
jaylco: This should be a href to the homepage; the removed `<a href="/" title="Go to the Civil Service Learning homepage" id="proposition-name">Civil Service Learning</a>` from below

jaylco (approved):

PR #1105, opened 2026-07-13
LC-4279: Search from nav bar

SavvasCO:


views/nunjucks/components/search/macro.njk
```diff
@@ -16,9 +16,6 @@
             <div class="column-one-half ">
                 {{ params.headerHtml | safe }}
             </div>
-            <div class="column-one-half ">
-                {{ searchBox(params.searchBox) }}
```
jaylco, line 20: Let's remove the `searchBox` macro if we aren't using it any more

views/nunjucks/components/search/macro.njk
```diff
@@ -16,9 +16,6 @@
             <div class="column-one-half ">
                 {{ params.headerHtml | safe }}
```
jaylco: Can we pull the headers out of the `search/macro.njk` and just have them render normally in the homepage, search and catalogue please. We can then remove the `search` macro altogether

views/assets/styles/components/_header.scss
```diff
@@ -228,4 +234,27 @@ a.js-header-toggle.menu.js-hidden {
...
+
+.search-form-container{
+	form {
+		display: flex;
+		align-items: center;
+		padding: 0 15px;
+
+		margin-top: 15px;
+
+		@include media(desktop) {
```
jaylco, line 247: Can we add `padding: 0` here as well, so that it sits flush against the right side of the navbar in desktop

views/assets/styles/components/_header.scss
```diff
@@ -228,4 +234,27 @@ a.js-header-toggle.menu.js-hidden {
...
+			margin-top: 0;
+		}
+
+		input.search-box__input{
+			width: 100%;
```
jaylco: Can we fix the width of this please rather than a percent? It's around 205px when I view it in devtools. It currently doesn't display correctly in Firefox due to this so we can assume that other browsers could be affected

jaylco (changes requested):

jaylco (approved):

